PR #1, opened 2020-04-03
[no ticket] add GitHub actions ci

andy7i:
Adds CI tests, linting


.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+      - name: Build with Gradle
+        run: ./gradlew build -x minniekenny # skip minniekenny
```
marctalbott, line 26: I don't know much about minniekenny, but can we run it here to detect if any secrets were checked in by someone who didn't run unit tests before they pushed their branch?

andy7i: discussed this with Doug and Will and the conclusion we came to is:
1) A secret that is committed and pushed to remote is already exposed.
2) Devops already has a scan of exposed secrets, so this would duplicate that.
3) Minnie-kenny is our best effort approach.

PR #2, opened 2020-04-29
WIP [CA-751] First implementation of CRL tests

marctalbott:
There are a _lot_ of changes in this PR, most of which are because the linter made a lot of changes. I broke this branch off of Doug's `cadence` branch, which means there are even more changes that I had nothing to do with. What I _am_ looking for feedback on is anything in the `src/test` directory and I suppose the changes I made to `GoogleCloudStorage.java` although I think those changes may be less controversial.


src/main/java/bio/terra/cloudres/google/storage/GoogleCloudStorage.java
```diff
@@ -0,0 +1,60 @@
...
+  public GoogleCloudStorage(Credentials credentials) {
+    this.storage = StorageOptions.newBuilder().setCredentials(credentials).build().getService();
+    this.credentials = credentials;
+  }
+
+  public boolean deleteBucketRaw(String bucketName) {
+    return storage.delete(bucketName);
+  }
+
+  public void setDeleteLifecycleRaw(String bucketName, Integer ageDays) {
```
wchamber: I don't think we should have this function. I think it should just be
`public void update(Bucket bucket) {...}`
clients can set lifecycle rules on Bucket without us.

src/test/java/bio/terra/cloudres/google/storage/GoogleCloudStorageTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+import org.springframework.boot.web.client.RestTemplateBuilder;
+import bio.terra.cloudres.util.GPAllocService;
+import bio.terra.cloudres.util.Project;
+
+@Tag("integration")
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = Main.class)
```
wchamber, line 25: what does this do? Seems funny to use anything from Main for our testing.

src/test/java/bio/terra/cloudres/util/Project.java
```diff
@@ -0,0 +1,12 @@
+package bio.terra.cloudres.util;
+
+import java.io.Serializable;
+
+public class Project implements Serializable {
```
wchamber, line 5: what's up with this class?

src/test/java/bio/terra/cloudres/google/storage/GoogleCloudStorageTest.java
```diff
@@ -0,0 +1,53 @@
...
+@SpringBootTest
+public class GoogleCloudStorageTest {
+
+  @Test
+  public void shouldCreateBuckets() throws Exception {
+    // todo: how do we get service account credentials in other integration/automation tests
+    String saKeyFile = "/Users/mtalbott/Downloads/mtalbott-test-billing-project-ef361ecead75.json";
```
wchamber, line 32: Yeah, this looks about like where we want to start in terms of the test fixture? I guess I don't understand CA-751 vs CA-750.
So we need to abstract out where the project and SA with credentials on the project comes from, get the values from a configuration file, which is populated... somehow by our github action?

src/test/java/bio/terra/cloudres/util/Project.java
marctalbott: I used this to represent the response that GPAlloc gives when requesting a project so we can serialize the response into a `Project`

src/test/java/bio/terra/cloudres/google/storage/GoogleCloudStorageTest.java
yonghaoy: Not sure if this helpful, seems that is how workbenlib work:
https://github.com/broadinstitute/workbench-libs/blob/develop/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/UserPool.scala#L29

src/main/java/bio/terra/cloudres/app/Main.java
```diff
@@ -0,0 +1,14 @@
+package bio.terra.cloudres.app;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
+import org.springframework.context.annotation.ComponentScan;
+
+@SpringBootApplication(exclude = {DataSourceAutoConfiguration.class})
```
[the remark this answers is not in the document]
marctalbott: I am happy to change any Spring things around, I am far from a Spring expert. I added Spring into the project in the first place so I could make requests to GPAlloc in my test, and adding in this Main.java file and all the other Spring stuff was just in the pursuit of making all the tests work correctly.

src/test/java/bio/terra/cloudres/google/storage/GoogleCloudStorageTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+@Tag("integration")
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = Main.class)
+@SpringBootTest
+public class GoogleCloudStorageTest {
+
+  @Test
+  public void shouldCreateBuckets() throws Exception {
+    // todo: how do we get service account credentials in other integration/automation tests
```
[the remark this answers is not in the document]
marctalbott: I just haven't spent much time thinking about how to render and read in config values yet beyond that it needs to happen. I assume we want a HOCON config file that's templated and can pull from vault and all that. I don't yet know the Java way of reading that config value, but I can't imagine that it's too difficult.

As for the question posed in the comment on this line, "how do we get service account credentials in other integration/automation tests", I think the answer for how we get credentials and permissions for the CRL tests will differ slightly from how we have done this previously since we won't be able to grant permissions on the GPAlloc projects using Rawls for these tests.

I think we need a new service account whose credentials will be stored in Vault and rendered/copied onto whatever machine is running these tests. That service account will need to be able to grant permissions on gpalloc-granted projects which all live in the test.firecloud.org or quality.firecloud.org GCP organizations. This service account would be responsible for requesting a project from gpalloc and then granting the required permissions to that project to the test user(s) for the duration of the test so that they can deal with the cloud resources in an existing project. Once the test is over, this SA would release the gpalloc project after revoking the previously-granted permissions.

That's how I've been thinking about it lately, but I am open to suggestions and further discussion

PR #4, opened 2020-05-28
CA-813 Add an integration test that reads Google Service Account Credentials.

wchamber:
Set up the github action to render the service account file from Vault and upload gradle test reports.
Add tasks to build.gradle to allow unit and integration tests to run separately.
Add a  integration test to be exercised (adding gradle dependencies as needed).
Add a stub section to the README about running integration tests locally.

I verified locally that the tests fail when they should.


src/test/java/bio/terra/cloudres/HelloTest.java
```diff
@@ -0,0 +1,14 @@
+package bio.terra.cloudres;
```
yonghaoy, line 1: I deleted the Hello class in my another PR, this is not needed?

src/test/java/bio/terra/cloudres/common/IntegrationCredentials.java
```diff
@@ -0,0 +1,22 @@
+package bio.terra.cloudres.common;
```
yonghaoy, line 1: does this should be under testing/ directory?

src/test/java/bio/terra/cloudres/common/IntegrationCredentialsTest.java
```diff
@@ -0,0 +1,13 @@
+package bio.terra.cloudres.common;
+
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+@Tag("integration")
+public class IntegrationCredentialsTest {
+  @Test
```
yonghaoy, line 8: I thought the general practice is let the test throws Exception, but we can add
try {
      
   } catch(yourException e) {
      fail("Should not have thrown any exception");
   }

.github/workflows/build-and-test.yaml
```diff
@@ -8,14 +8,38 @@ on:
...
+          echo ::set-output name=vault-token::$VAULT_TOKEN
+          echo ::add-mask::$VAULT_TOKEN
+      - name: Get service account credentials from Vault
+        id: vault-secret-step
+        run: |
+          docker run --rm --cap-add IPC_LOCK \
+            -e "VAULT_TOKEN=${{ steps.vault-token-step.outputs.vault-token }}" \
+            -e "VAULT_ADDR=${VAULT_ADDR}" \
+            vault:1.1.0 \
+            vault read -format json secret/dsde/terra/crl-test/default/service-account.json \
```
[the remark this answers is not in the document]
wchamber: good idea, thanks.

src/test/java/bio/terra/cloudres/HelloTest.java
wchamber: Yep, thanks. just needed a unit test to run.

src/test/java/bio/terra/cloudres/common/IntegrationCredentials.java
wchamber: done.

src/test/java/bio/terra/cloudres/common/IntegrationCredentialsTest.java
wchamber: I like that that makes it more explicit, done. Do you have a source? Not sure I've seen that before.

yonghaoy: Forgot where I saw the documented source, that might come from some someone else's recommendation during code review before.

Just found an google internal  discussion: https://yaqs.corp.google.com/eng/q/5403980279709696#a6708475597029376

PR #5, opened 2020-05-29
[COW] Create Google project including tracing, logging, metrics

yonghaoy:


src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -0,0 +1,82 @@
...
+import io.opencensus.trace.Tracer;
+import io.opencensus.trace.Tracing;
+import java.time.Duration;
+import java.util.OptionalInt;
+import java.util.function.Supplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Annotates executing cloud operations with logs, traces, and metrics to record what happens. */
+public class OperationAnnotator {
```
gpolumbo-broad, line 19: My preference would be to call this class something else like, `OperationRunner` or `OperationExecutor` or `CloudCaller` or something like that.   To me, the primary purpose of this class is that it performs the cloud operation.  The way it's named now, it sort of implies to me that it's primary purpose is to annotate.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -0,0 +1,82 @@
...
+   * @param request: the formatted request, this is used for logging.
+   */
+  public <R> R executeGoogleCall(Supplier<R> googleCall, CloudOperation operation, String request) {
+    OptionalInt errorCode = OptionalInt.empty();
+    R response = null;
+
+    Stopwatch stopwatch = Stopwatch.createStarted();
+    try (Scope ss = tracer.spanBuilder(operation.name()).startScopedSpan()) {
+      // Record the Cloud API usage.
+      ;
```
gpolumbo-broad, line 42: dangling `;`

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -0,0 +1,82 @@
...
+      } finally {
+        logEvent(
+            /*logger=*/ logger,
+            /*traceId=*/ tracer.getCurrentSpan().getContext().getTraceId(),
+            /* operation=*/ CloudOperation.GOOGLE_CREATE_PROJECT,
+            /* clientName=*/ clientConfig.getClientName(),
+            /* request=*/ request,
+            /* response=*/ JsonConverter.convert(response),
+            /* response=*/ errorCode);
+        recordLatency(stopwatch.stop().elapsed(), operation);
```
gpolumbo-broad, line 60: Is this the right location for recording latency?  Where we have it now, it's capturing some metrics gathering and logging events.  I'm not sure if we want the stop watch specifically around just the `googleCall.get()`?

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
wchamber: I'll bite. It is to do the annotations. We don't need this class to invoke the Supplier, we need the Supplier separated to do the annotations happening in this class.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  /**
+   * Executes the Google call.
+   *
+   * @param googleCall: the google call to make
+   * @param operation: the {@link CloudOperation}
+   * @param request: the formatted request, this is used for logging.
+   */
+  public <R> R executeGoogleCall(Supplier<R> googleCall, CloudOperation operation, String request) {
```
gpolumbo-broad, line 35: In the `logEvent` method, we do some json appending to the `request`.  At the very least we should comment that this parameter needs to be valid json.  If we can more strictly enforce this type requirement, that would be even better.

wchamber, line 35: I think we might end up with a CloudCall interface which includes a way to serialize the request for logging, and an execute function that returns a result that knows how to get the underlying value for logging. But let's just stick a TODO on here.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: done

src/main/java/bio/terra/cloudres/util/JsonConverter.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Util class to convert cloud resources to Json format
+ *
+ * <p>CRL want to log full request aid for debugging. Added this to help us better log things in a
+ * good format
+ */
+public class JsonConverter {
+  /** Converts a generic type to json */
+  public static <R> String convert(R object) {
```
wchamber, line 17: this is a dangerous overload. (Sorry if I missed some comment on why it exists). This implies we can do the right thing with most types. I don't think that's true. If it is, then why do we need the specific Project and ProjectInfo overloads.
But I'll settle for a TODO to think about if we want this function in the general logging PR to follow.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: oh, right, moved to try and catch. Thanks!

src/main/java/bio/terra/cloudres/util/JsonConverter.java
yonghaoy: Sorry, forgot to remove the bottom two things. I tested it out, it works with Project, and projectInfo. I think for google's one it will work as most of them are passed from proto. 
The ideal way in my mind is to have a JsonTypeAdaper factory. But I think we can added as we need?

yonghaoy: Sorry, ,missed one point. The reason why it exist is because we are doing most things in the annotator and it doesn't know the requuest/respons type.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: Changed, now it won't expect any formtted stuff. It will handle all JsonConvert in log metrics.

yonghaoy:
TODO left

1. Make OperationAnnotator takes logger from Cow
2. Skip response in log
3. Log error message from exception
4. Request might be multiple and Cow requestFormatter pass to 
5. Latency in log
6: move start stopwatch closer to api call

7: Metrics exporter 
8: ./gradlew spotlessApply plus javadoc

9: Readme

PR #6, opened 2020-05-29
Refactor CRL based on feedback around logging stuff

yonghaoy:
This PR covers:

- Introduce interface CowOperation which would be the only parameter passed into OperationAnnotator
- Remove LogHelper since all are handled in OperationAnnotator
- Remove response from log
- Log error message and error code
- Move stop watch start/stop close to the cloud call
-


src/main/java/bio/terra/cloudres/common/CowOperation.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * interface for a single Cloud Operation
+ *
+ * <p> We expect each Cow implements its own CowOperations includes:
+ * <ul>
+ *     <li> The {@code CloudOperation}
+ *     <li> how to execute this operation
+ *     <li> how to serialize Request
+ * </ul>
```
wchamber, line 11: nit: I think removing this section and commenting on each of the functions instead is a little better for documentation.

src/main/java/bio/terra/cloudres/common/CowOperation.java
```diff
@@ -0,0 +1,20 @@
+package bio.terra.cloudres.common;
+
+/**
+ * interface for a single Cloud Operation
```
wchamber, line 4: comment suggestion: Abstraction about what {@link OperationAnnotator} needs to know about a single cloud operation to annotate what happens on execution.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -1,62 +1,71 @@
...
   private static final Tracer tracer = Tracing.getTracer();
-  private final Logger logger = LoggerFactory.getLogger(OperationAnnotator.class);
+  /**
+   * Fake HTTP status code value for errors that are not HTTP status errors. Useful for including
+   * non-HTTP status errors in a single metric.
+   */
+  public static final int GENERIC_UNKNOWN_ERROR_CODE = 1;
+
   private final ClientConfig clientConfig;
+  private final Logger logger;
```
wchamber, line 31: Worth a comment on why we pass in a logger. suggestion:
/** We inject a Logger to allow how logs are made to be controlled by the COWs. */

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -1,62 +1,71 @@
...
 
   /**
    * Executes the Google call.
    *
-   * @param googleCall: the google call to make
-   * @param operation: the {@link CloudOperation}
-   * @param request: the request of this execute
+   * @param cowOperation: the {@link CowOperation} whichs contains all necessary information to
+   *     execute this call.
+   * @return the response from cloud
```
wchamber, line 43: nit: the result of executing the {@code cowOperation}

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -1,62 +1,71 @@
...
+  private final Logger logger;
 
-  public OperationAnnotator(ClientConfig clientConfig) {
+  public OperationAnnotator(ClientConfig clientConfig, Logger logger) {
     this.clientConfig = clientConfig;
+    this.logger = logger;
   }
 
   /**
    * Executes the Google call.
```
wchamber, line 39: nit: maybe 'Executes the CowOperation.'

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -1,62 +1,71 @@
...
       } finally {
         logEvent(
-            /*logger=*/ logger,
             /*traceId=*/ tracer.getCurrentSpan().getContext().getTraceId(),
-            /* operation=*/ CloudOperation.GOOGLE_CREATE_PROJECT,
-            /* clientName=*/ clientConfig.getClientName(),
-            /* requestFormatter=*/ request,
-            response,
-            /* response=*/ errorCode);
+            /* operation=*/ cloudOperation,
```
wchamber, line 66: nit: I think we can remove the /* param= */ comments now - they each have distinct types that are decently descriptive.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -1,62 +1,71 @@
...
-        response = googleCall.get();
-        recordLatency(stopwatch.stop().elapsed(), operation);
+        R response = cowOperation.execute();
+        recordLatency(stopwatch.stop().elapsed(), cloudOperation);
         return response;
       } catch (Exception e) {
-        recordLatency(stopwatch.stop().elapsed(), operation);
-        errorCode = getHttpErrorCode(e);
-        recordErrors(errorCode, operation);
+        recordLatency(stopwatch.stop().elapsed(), cloudOperation);
```
wchamber, line 59: hmmm, we could also recordLatency and recordErrors (if exception is present) in the 'finally'. WDYT?

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -1,62 +1,71 @@
...
+   * @param cowOperation: the {@link CowOperation} whichs contains all necessary information to
+   *     execute this call.
+   * @return the response from cloud
    */
-  public <R, T> R executeGoogleCall(Supplier<R> googleCall, CloudOperation operation, T request) {
-    OptionalInt errorCode = OptionalInt.empty();
-    R response = null;
+  public <R> R executeGoogleCall(CowOperation<R> cowOperation) {
+    CloudOperation cloudOperation = cowOperation.getCloudOperation();
+    Optional<Exception> exception = Optional.empty();
```
wchamber, line 47: nit: how about `executionException` or `operationException` to add a little more about what we're using this for to the name.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -78,4 +87,45 @@ private OptionalInt getHttpErrorCode(Exception e) {
         ? OptionalInt.of(((BaseHttpServiceException) e).getCode())
         : OptionalInt.empty();
   }
+
+  /**
+   * Logs cloud calls. This should be in debug level.
```
wchamber, line 92: nit: if we're saying it should be in debug level, we should say why it should be in debug level.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -78,4 +87,45 @@ private OptionalInt getHttpErrorCode(Exception e) {
...
+   *
+   * @param traceId the traceId where log happens
+   * @param operation the operation to log.
+   * @param request the request of the log
+   * @param exception the exception to log. Optional, only presents when exception happens.
+   */
+  @VisibleForTesting
+  void logEvent(
+      TraceId traceId, CloudOperation operation, String request, Optional<Exception> exception) {
```
wchamber, line 101: (if we change the variable name for the Optional<Exception> above, do it here too)

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -78,4 +87,45 @@ private OptionalInt getHttpErrorCode(Exception e) {
...
+    if (logger.isDebugEnabled()) {
```
wchamber, line 102: nit: if we're not going to do anything ever if debug isn't enabled, any early return makes that more clear.
```
if (!logger.isDebugEnabled())  {
  return;
}
// do the actual work of the function
```

src/main/java/bio/terra/cloudres/util/JsonConverter.java
```diff
@@ -11,22 +12,16 @@
...
    *
-   * @param object: the object to convert
+   * @param projectInfo: the projectInfo to convert
    * @return the formatted Json in String
    */
-  public static <R> String convert(R object) {
-    if (object == null) {
-      return null;
-    }
+  public static String convert(ProjectInfo projectInfo) {
```
wchamber, line 22: I would say don't even have this file anymore. It's Cow specific (and very short in this case). We can inline it in the ResourceManagerCow, or have the function in a utils file for that resourcemanager package.
Either way, we don't want all of the different cloud resources to get imported into a common package like utils, or every package will have a dependency on every other package.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -1,62 +1,71 @@
...
 
 /** Annotates executing cloud operations with logs, traces, and metrics to record what happens. */
 public class OperationAnnotator {
   private static final Tracer tracer = Tracing.getTracer();
-  private final Logger logger = LoggerFactory.getLogger(OperationAnnotator.class);
+  /**
+   * Fake HTTP status code value for errors that are not HTTP status errors. Useful for including
+   * non-HTTP status errors in a single metric.
+   */
+  public static final int GENERIC_UNKNOWN_ERROR_CODE = 1;
```
wchamber, line 28: I still think this belongs in MetricsHelper. This variable isn't even used in this class.
How MetricsHelper serializes unknown HTTP status codes is its implementation details. Logically, we want to be able to record errors whether or not we know the HTTP status code. That "maybe we don't know it" is a part of the interface to MetricsHelper.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -78,4 +87,45 @@ private OptionalInt getHttpErrorCode(Exception e) {
...
+      jsonMap.put("clientName:", clientConfig.getClientName());
+
+      String jsonString = JsonConverter.convert(jsonMap);
+
+      // Now append the already formatted request & exception.
+      Gson gson = new Gson();
+      JsonObject jsonObject = gson.fromJson(jsonString, JsonObject.class);
+
+      // If exception presents
+      if (exception.isPresent()) {
```
wchamber, line 115: nit: consider making a helper function to return the exceptionMap.

src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
```diff
@@ -2,47 +2,66 @@
...
+  private static final String PROJECT_INFO_STRING =
+      "{\"name\":\"myProj\",\"projectId\":\"project-id\",\"labels\":{\"k1\":\"v1\",\"k2\":\"v2\"}}";
+  private static final String TRACE_ID = "1234567890123456";
+  private static final String ERROR_MESSAGE = "error!";
+  private static final String FORMATTED_EXCEPTION =
+      "\"exception:\":{\"message\":\"error!\",\"errorCode\":\"404\"},";
+  private static final String EXPECTED_LOG_PREFIX =
+      "{\"traceId:\":\"TraceId{traceId=31323334353637383930313233343536}\",\"operation:\":\"GOOGLE_CREATE_PROJECT\",\"clientName:\":\"TestClient\",";
+
+  private static final ResourceManagerException RM_EXCEPTION =
```
wchamber, line 34: Worth a comment that we're using ResourceManagerException as an example of a BaseHttpServiceException, or even make our own subclass in this test to make that clear.

src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
```diff
@@ -2,47 +2,66 @@
...
+      new ResourceManagerException(404, ERROR_MESSAGE);
 
   private OperationAnnotator operationAnnotator;
   private ClientConfig clientConfig;
+  ArgumentCaptor<String> logArgument = ArgumentCaptor.forClass(String.class);
+  @Mock private Logger logger = mock(Logger.class);
+  @Mock private CowOperation mockCowOperation = mock(CowOperation.class);
```
wchamber, line 41: Prefer making your own real CowOperation implementations to mocking it. There's nothing to stop this test from making it's own implementations. Mocks are the testing tool of last resort. go/mocks

It's fine to add helpers to make the CowOperations you want read fluently in the test code. So, have a 'ExceptionCowOperation(RuntimeException e)' and a 'SuccessfulCowOperation(String executionResult)'.

src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
```diff
@@ -2,47 +2,66 @@
...
 
   @BeforeEach
   public void setUp() throws Exception {
     clientConfig = ClientConfig.Builder.newBuilder().setClient(CLIENT).build();
-    operationAnnotator = new OperationAnnotator(clientConfig);
+    operationAnnotator = new OperationAnnotator(clientConfig, logger);
```
wchamber, line 46: nit: Just initalize operationAnnotator on its declaration line (37). Inline variable declarations are even simpler than setUp generally and should be preferred when they're short.
For most of the tests, a default clientConfig and a real logger are fine. For the test that cares about what gets logged, only use a mock logger in that test.

src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
```diff
@@ -86,4 +105,92 @@ public void testExecuteGoogleCloudCall_withException() throws Exception {
...
+    operationAnnotator.logEvent(
+        TraceId.fromBytes(TRACE_ID.getBytes()),
+        CloudOperation.GOOGLE_CREATE_PROJECT,
+        PROJECT_INFO_STRING,
+        Optional.empty());
+
+    // Expected result in Json format
+    verify(logger).debug(logArgument.capture());
+    assertEquals(
+        EXPECTED_LOG_PREFIX + "\"request:\":" + PROJECT_INFO_STRING + "}", logArgument.getValue());
```
wchamber, line 146: eh, I'm not a huge fan of the string manipulation to match the expected json, but there's not a lot of good options. One alternative would be to make the EXPECTED_* objects json objects and concatenate them as needed with the json class apis, then assert they serialize the same. Not a strong opinion here, I'm not a huge fan of the test because they detect changes.
Maybe a better way would be to use a @VisibleForTesting helper function in OperationAnnotator that returns the Json object. Then, rather than messing with mocking logs, we test that helper function directly.
so in operation annotator
```
private void logEvent(....) {
  logger.debug(logEventToJson(...).toString())
}

@VisibleForTesting
static JsonObject logEventToJson(...) {
   // build up JsonObject
}
```

src/test/java/bio/terra/cloudres/testing/MetricsTestUtil.java
```diff
@@ -22,7 +22,7 @@
       Arrays.asList(
           TagValue.create(CLIENT),
           TagValue.create(CloudOperation.GOOGLE_CREATE_PROJECT.name()),
-          TagValue.create("0"));
+          TagValue.create("404"));
```
wchamber, line 25: That this is changing means tests are depending too deeply on what this is defined as. Tests should not be asserting about properties of public test quantities in ways that are not specified by their names.
 Either make it more explicit that this is a ERROR_COUNT_404 or inline this in the test that care about the value itself.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -78,4 +87,45 @@ private OptionalInt getHttpErrorCode(Exception e) {
...
+        exceptionMap.put("message", exception.get().getMessage());
+        exceptionMap.put(
+            "errorCode",
+            String.valueOf(getHttpErrorCode(exception.get()).orElse(GENERIC_UNKNOWN_ERROR_CODE)));
+
+        JsonConverter.appendFormattedString(
+            jsonObject, "exception:", JsonConverter.convert(exceptionMap));
+      }
+
+      JsonConverter.appendFormattedString(jsonObject, "request:", request);
```
wchamber, line 127: I don't find our JsonConverter functions particularly helpful. What's wrong with manipulating the JsonObjects directly?

src/main/java/bio/terra/cloudres/common/CowOperation.java
yonghaoy: done

src/main/java/bio/terra/cloudres/common/CowOperation.java
yonghaoy: done

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: renamed to executeCowOperation. execute() sounds similar to the method in cloudOperation.

src/main/java/bio/terra/cloudres/google/crm/ResourceManagerCow.java
```diff
@@ -34,9 +40,29 @@ public ResourceManagerCow(
...
+
+    @Override
+    public Project execute() {
+      return resourceManager.create(projectInfo);
+    }
+
+    @Override
+    public String serializeRequest() {
+      return JsonConverter.convert(projectInfo);
+    }
```
dvoet, line 66: I like the idea of using an interface but this is really verbose. How about an anonymous class:

```
    return operationAnnotator.executeGoogleCall(new CowOperation<Project>() {
      @Override public CloudOperation getCloudOperation() { return CloudOperation.GOOGLE_CREATE_PROJECT; }
      @Override public Project execute() { return resourceManager.create(projectInfo); }
      @Override public String serializeRequest() { return JsonConverter.convert(projectInfo); }
    });
```

Maybe even better would be to use lamdas:
```
    return operationAnnotator.executeGoogleCall(
      CloudOperation.GOOGLE_CREATE_PROJECT,
      () -> resourceManager.create(projectInfo),
      () -> JsonConverter.convert(projectInfo)
    );
```
where there exists:
```
  public interface CowExecute<R> {
    R execute();
  }

  public interface CowSerialize {
    String serializeRequest();
  }

  public <R> R executeGoogleCall(CloudOperation op, CowExecute<R> ex, CowSerialize seralizer) {
```

yonghaoy: Good point! Changed to anonymous class.
Seems I can not use Lambda, because "Lambda expressions can only be used to implement functional interfaces, which are interfaces having a single abstract method. An interface with two abstract methods can't be implemented by a lambda expression."

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: Done. But I usually have the param= block if there are too many parameters need to pass.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: That would make the code more complicated than it is now. We will have if(exception.isPresent) in finally block. I think recordLatency should be in try and catch to make sure that line is close to the cloud call(e.g., there is one more line happens in try block after the cloud call. we may have to move recordLatency to catch again). Let me know if I understand this comment corretcly. Thanks!

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: added some comments. Not sure if that explains, let me know if the javadoc sounds not right. Thanks

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: removd this function

src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
yonghaoy: Done. Not sure if that is what you want. PTAL, thanks

src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
yonghaoy: Added some comments. Still use the RM one because I didn't see too many benefits of creating our own one.. is that okay?

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: This code is now used for logg as well(see getExceptionMap). If an internal error happens, we also log the error code, and that would make the log consistent with metrics, and easier to lookup?

src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
yonghaoy: We are not teting "he string manipulation to match the expected json", what we want to test is the arguments passed into in logger is correct, i.e., the String formatted converted from Json.
I don't think one more helper logEventToJson tests better than what we have now, so I have a long javadoc in Json format(pased from jsonformatter).

If you think we should test the jsonOject we creatd, I can copied that logic and create a jsonOjbect. One thing I propose
If we want to test JsonObject, I feels in test, we are rewritting most of the code in logEvent: 
    Map<String, String> jsonMap = new LinkedHashMap<>();
    jsonMap.put("traceId:", traceid);
    jsonMap.put("operation:", operation);
    jsonMap.put("clientName:", operation);
....

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
wchamber: You understood correctly, sounds fine to me to keep as-is.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
wchamber: I don't think the metrics and json log need to serialize in the same way. For the JSON log, we can just omit the errorCode field when there's no exception - that's not really hard to search for and is more JSON idiomatic. For Metrics, we don't have a "missing" or "not present" concept,so we need the unknown code.

src/main/java/bio/terra/cloudres/google/crm/ResourceManagerCow.java
wchamber: Due to google-java-format, the inline anonymous CowOperation class is not really any less verbose - each function gets expanded over 3 lines. I think it's fine to use the inline one here, don't have a strong feeling.
I don't mind having 3 different arguments passed in with interfaces instead, but I kind of like tying things up in a single interface. I'm leaning a little towards the single class, but could be persuaded. Having the single CowOperation interface with separate implementing classes makes them easier to re-use, but it's more Java verbosity. But Java verbosity is just annoying, it's not really hard to work with.
We might re-use the separate classes because Google has multiple ways to do things like create Blobs: https://googleapis.dev/java/google-cloud-storage/latest/com/google/cloud/storage/Storage.html#create-com.google.cloud.storage.BlobInfo-byte:A-int-int-com.google.cloud.storage.Storage.BlobTargetOption...-
https://googleapis.dev/java/google-cloud-storage/latest/com/google/cloud/storage/Bucket.html#create-java.lang.String-byte:A-com.google.cloud.storage.Bucket.BlobTargetOption...-

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -82,50 +83,62 @@ private void recordLatency(Duration duration, CloudOperation operation) {
...
+    jsonMap.put("traceId:", traceId.toString());
+    jsonMap.put("operation:", operation.name());
+    jsonMap.put("clientName:", clientConfig.getClientName());
+
+    String jsonString = gson.toJson(jsonMap);
 
-      JsonConverter.appendFormattedString(jsonObject, "request:", request);
-      logger.debug(jsonObject.toString());
+    // Now append the already formatted request & exception.
+    JsonObject jsonObject = gson.fromJson(jsonString, JsonObject.class);
```
wchamber: this goes back and forth between Strings, JsonObjects, and Maps. What if we just use JsonObject?
```
JsonObject logEntry = new JsonObject();
logEntry.addProperty("traceId", traceId.toString());
logEntry.addProperty("operation", operation.name());
...
```
ditto for the exception - change getExceptionMap -> createExceptionEntry (or a better name) returning a JsonObject.
Then
`logEntry.addProperty("exception", createExceptionEntry(executionException.get()));`

src/main/java/bio/terra/cloudres/google/crm/ResourceManagerCow.java
```diff
@@ -40,29 +40,33 @@ public ResourceManagerCow(
...
-    public String serializeRequest() {
-      return JsonConverter.convert(projectInfo);
-    }
+  /**
+   * Converts {@link ProjectInfo} to Json formatted String
+   *
+   * @param projectInfo: the projectInfo to convert
+   * @return the formatted Json in String
+   */
+  private static String convert(ProjectInfo projectInfo) {
```
wchamber, line 68: I think it's fine to inline this function (and I don't think it needs javadoc)

src/main/java/bio/terra/cloudres/util/JsonConverter.java
```diff
@@ -1,51 +1,9 @@
...
-   * @param value the value to append
-   * @return The new {@link JsonObject} after append
-   */
-  public static JsonObject appendFormattedString(JsonObject jsonObject, String key, String value) {
-    Gson gson = new Gson();
-    jsonObject.add(key, gson.fromJson(value, JsonObject.class));
-    return jsonObject;
-  }
-}
+public class JsonConverter {}
```
wchamber, line 9: remove file for empty class

src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
wchamber: Yep, that's good with comment.

src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
```diff
@@ -31,37 +32,41 @@
...
+  /** use the {@link ResourceManagerException} as an example of a BaseHttpServiceException. */
   private static final ResourceManagerException RM_EXCEPTION =
       new ResourceManagerException(404, ERROR_MESSAGE);
 
-  private OperationAnnotator operationAnnotator;
-  private ClientConfig clientConfig;
+  private static final Supplier FAILED_COW_EXECUTE_SUPPLIER =
+      () -> {
+        throw RM_EXCEPTION;
+      };
```
wchamber, line 42: if this is only used once, maye just inline it.

src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
wchamber: sounds good.

src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
```diff
@@ -175,22 +177,46 @@ public void testLogEvent_nullResponse() throws Exception {
...
-    when(logger.isDebugEnabled()).thenReturn(false);
+    operationAnnotator = new OperationAnnotator(clientConfig, mockLogger);
+    when(mockLogger.isDebugEnabled()).thenReturn(false);
 
     operationAnnotator.logEvent(
         TraceId.fromBytes(TRACE_ID.getBytes()),
         CloudOperation.GOOGLE_CREATE_PROJECT,
         PROJECT_INFO_STRING,
         Optional.of(RM_EXCEPTION));
     // Expected result in Json format
```
wchamber, line 196: no expected result in this case

src/main/java/bio/terra/cloudres/google/crm/ResourceManagerCow.java
yonghaoy: Just saw the blob may need different execute() methods. In that case, we may want our creatBlobOperation class has constructor to pass the exceute function. In the createProject case, if we want to resume some code for delete project or update/list project(assume they all require the projectInfo, which is the deplicate part when convert them): we can have seperate class like:

class RMCowOperation  implements CowOperation<Project> {
    ProjectInfo projectinfo;
    CloudOperation cloudOperation;
    Supplier executeFunc

 RMCowOperation(CloudOperation, Supplier executeFunc, ProjectInfo projectinfo) {}
          @Override
          public String serializeRequest() {
            return convert(projectInfo);
          }
}

But not sure if that is valuable?

src/main/java/bio/terra/cloudres/google/crm/ResourceManagerCow.java
yonghaoy: I think this will be reused when we have other cloud calls similar to this one: like delete/update project?

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: done

src/main/java/bio/terra/cloudres/google/crm/ResourceManagerCow.java
wchamber: I don't think I want an abstract RMCowOperation yet. I'd say wait until we need it, see what looks good then.

src/main/java/bio/terra/cloudres/google/crm/ResourceManagerCow.java
wchamber: good point. keep it separate.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -104,22 +104,12 @@ void logEvent(
     }
     Gson gson = new Gson();
 
-    Map<String, String> jsonMap = new LinkedHashMap<>();
-    jsonMap.put("traceId:", traceId.toString());
-    jsonMap.put("operation:", operation.name());
-    jsonMap.put("clientName:", clientConfig.getClientName());
+    JsonObject jsonObject = new JsonObject();
```
wchamber, line 107: make the variable name a little more descriptive about what it's for, e.g. `logEntry` instead of `jsonObject`

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -132,13 +122,19 @@ private OptionalInt getHttpErrorCode(Exception e) {
         : OptionalInt.empty();
   }
 
-  private Map<String, String> getExceptionMap(Exception executionException) {
+  private void createExceptionEntry(Optional<Exception> executionException, JsonObject jsonObject) {
```
wchamber, line 125: prefer return values to output parameters. We can return a JsonObject, or an Optional<JsonObject> and handle that it goes under the "exception" field in the calling function.
e.g. in logEvent
`executionException.ifPresent(e -> logEntry.add("exception", createExceptionEntry(e));`

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -104,22 +104,12 @@ void logEvent(
     }
     Gson gson = new Gson();
```
wchamber, line 105: nit: move this down to right before it's used.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
wchamber: still outstanding

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -73,9 +83,58 @@ private void recordLatency(Duration duration, CloudOperation operation) {
...
+    Gson gson = new Gson();
+
+    JsonObject jsonObject = new JsonObject();
+    jsonObject.addProperty("traceId:", traceId.toString());
+    jsonObject.addProperty("operation:", operation.name());
+    jsonObject.addProperty("clientName:", clientConfig.getClientName());
+
+    createExceptionEntry(executionException, jsonObject);
+
+    jsonObject.add("request:", gson.fromJson(request, JsonObject.class));
```
wchamber, line 114: This is kind of funny. Do we think `request` should actually be a JsonObject, and `CowOperation` serializeRequest should return a JsonObjcet? Looks like we're coercing it.

src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
wchamber: yep, thanks.

src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
yonghaoy: this is also used in log tests below where we test if exception are logge correctly.

src/main/java/bio/terra/cloudres/google/crm/ResourceManagerCow.java
dvoet: I think the verbosity is more than just annoying. It hides the actual functionality each Cow operation is implementing making it harder maintain. I don't think there is much opportunity reuse any of the CowOperation implementations. I think reusing instances is a fallacy that will lead to incorrect logging, etc.

I think we should go with 3 arguments and 2 single function interfaces .

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -1,62 +1,72 @@
...
-        response = googleCall.get();
-        recordLatency(stopwatch.stop().elapsed(), operation);
+        R response = cowOperation.execute();
+        recordLatency(stopwatch.stop().elapsed(), cloudOperation);
         return response;
       } catch (Exception e) {
-        recordLatency(stopwatch.stop().elapsed(), operation);
-        errorCode = getHttpErrorCode(e);
-        recordErrors(errorCode, operation);
+        recordLatency(stopwatch.stop().elapsed(), cloudOperation);
```
gpolumbo-broad, line 60: Should/can we include a `success` or `error` indicator/flag/annotation when we call `recordLatency`?  In other words, when we're reviewing metrics, will we be able to differentiate latency records from successful Cloud Operations vs. Cloud Operations that threw an exception?  For debugging purposes, this feels like an important metric that I would want to be able to filter on.  This might be a comment best addressed in a separate PR.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: done

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: (1) but for the case where some unexpected exceptions happens throws Exception, do we want to log the e as well? In that case do we just want error message or with error code. 
(2) I also think the UNKNOWN_ERROR doesn't just belong to metrics if we want to keep the error code consistent everyewhere

wchamber: 1) just the error message, there is no error code.
2) Is there somewhere else besides the log & the metric? I still think a no recorded error code is clearer in the log than a  made up number. I think it's ok if we have fields in the log event that are only there when they make sense. We're not going to programatically access the metrics and the logs simultaneously, so sharing a number doesn't seem that important to me.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -73,9 +83,58 @@ private void recordLatency(Duration duration, CloudOperation operation) {
...
+   *
+   * <p>This log is for debug purpose when using CRL, so should be in debug level.
+   *
+   * @param traceId the traceId where log happens
+   * @param operation the operation to log.
+   * @param request the request of the log
+   * @param executionException the exception to log. Optional, only presents when exception happens.
+   */
+  @VisibleForTesting
+  void logEvent(
```
gpolumbo-broad, line 97: This method is pretty specific to Google Operations what with the `Gson` and other types from Google libs.  As a matter of fact, this class seems couple to Google Cloud Operations for now, yet the class name and package seem to imply it is generic, and I think that is concerning.  We don't know what the future will hold with other cloud providers, and that's fine, so I think my vote would be to maybe just change the package for this class to be something `google` specific instead of `common`.

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
gpolumbo-broad: I agree, I would rather `request` be a more specific type besides `String` and would not object to @wchamber 's suggestion above or adding a `CowOperation.toString()` or `CowOperation.toJson()` maybe?

src/main/java/bio/terra/cloudres/google/crm/ResourceManagerCow.java
```diff
@@ -34,9 +40,33 @@ public ResourceManagerCow(
    * @return the project being created
    */
   public Project createProject(ProjectInfo projectInfo) {
-    return operationAnnotator.executeGoogleCall(
-        () -> resourceManager.create(projectInfo),
-        CloudOperation.GOOGLE_CREATE_PROJECT,
-        projectInfo);
+    return operationAnnotator.executeCowOperation(
```
gpolumbo-broad, line 44: I've been out of Java land for several years and I'm not sure of the conventions, but I would prefer this be a "regular" defined class rather than having it be inlined here.

dvoet: https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/6#discussion_r433235115

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: Actually this is not google specify things, and this is intented to be generic.  Gson is a json format library. I compared this with gson vs jackson, and seems Gson is easier and cleaner to use. 
The only gcp things is the BaseHttpServiceException, and as we integrate more cloud providers we just need a if e isinstanceof OtherCloudException block

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: okay. added a comments(todo here). I think our metrics ticket will be a good fit for that as we furture improve metrics. I will try to pick that if that get prioritez.

src/main/java/bio/terra/cloudres/google/crm/ResourceManagerCow.java
yonghaoy: done

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -127,9 +127,8 @@ private JsonObject createExceptionEntry(Exception executionException) {
     // Nested Exception
     Map<String, String> exceptionMap = new LinkedHashMap<>();
     exceptionMap.put("message", executionException.getMessage());
-    exceptionMap.put(
-        "errorCode",
-        String.valueOf(getHttpErrorCode(executionException).orElse(GENERIC_UNKNOWN_ERROR_CODE)));
+    getHttpErrorCode(executionException)
+        .ifPresent(i -> exceptionMap.put("errorCode", String.valueOf(i)));
     return gson.fromJson(gson.toJson(exceptionMap), JsonObject.class);
```
wchamber: Direcly construct the JsonObject here too? like `logEntry` above

src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
```diff
@@ -86,4 +114,96 @@ public void testExecuteGoogleCloudCall_withException() throws Exception {
...
+   *
+   * <pre>{@code
+   * { "traceId:":"TraceId{traceId=31323334353637383930313233343536}",
+   *   "operation:":"GOOGLE_CREATE_PROJECT",
+   *   "clientName:":"test_client",
+   *   "errorCode:":"404",
+   *   "request:":{
+   *      "requestName":"request1"
+   *   },
+   *   "response:":null
```
wchamber, line 170: fix the colons inside of the json keys, e.g. `"traceId"` instead of `"traceId:"` Check if there are others in the file

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -73,9 +83,52 @@ private void recordLatency(Duration duration, CloudOperation operation) {
...
+    if (!logger.isDebugEnabled()) {
+      return;
+    }
+
+    JsonObject logEntry = new JsonObject();
+    logEntry.addProperty("traceId:", traceId.toString());
+    logEntry.addProperty("operation:", operation.name());
+    logEntry.addProperty("clientName:", clientConfig.getClientName());
+
+    executionException.ifPresent(e -> logEntry.add("exception:", createExceptionEntry(e)));
```
wchamber, line 111: I don't think we should have colons ':' in the property names.

src/main/java/bio/terra/cloudres/google/crm/ResourceManagerCow.java
```diff
@@ -34,9 +40,20 @@ public ResourceManagerCow(
...
+
+  /**
+   * Converts {@link ProjectInfo} to {@link JsonObject}
+   *
+   * @param projectInfo: the projectInfo to convert
+   * @return the formatted JsonObject
+   */
+  private static JsonObject convert(ProjectInfo projectInfo) {
+    Gson gson = new Gson();
+    return gson.fromJson(gson.toJson(projectInfo, ProjectInfo.class), JsonObject.class);
```
wchamber, line 57: does `return gson.toJsonTree(projectInfo, ProjectInfo.class)`; work instead? Seems like we should be able to get to a JsonObject without going through an intermediate object.

src/test/java/bio/terra/cloudres/testing/MetricsTestUtil.java
wchamber: rename to ERROR_COUNT_404?

src/main/java/bio/terra/cloudres/common/CowOperation.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+public class CowOperation {
+
+  /** How to execute this operation */
+  public interface CowExecute<R> {
+    R execute();
+  }
+
+  /** How to serialize Request */
+  public interface CowSerialize {
```
wchamber, line 17: I think we can instead define the interfaces directly on OperationAnnotator where they are used and don't need this class/file anymore.

src/main/java/bio/terra/cloudres/google/crm/ResourceManagerCow.java
yonghaoy: done

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: done

src/main/java/bio/terra/cloudres/common/CowOperation.java
yonghaoy: done

src/main/java/bio/terra/cloudres/google/crm/ResourceManagerCow.java
yonghaoy: that returns JsonElements. But it works as I can convert that to jsonObject:
.getAsJsonObject()

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: done

src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -73,9 +79,61 @@ private void recordLatency(Duration duration, CloudOperation operation) {
...
+    Gson gson = new Gson();
+    JsonObject exceptionEntry = new JsonObject();
+    exceptionEntry.addProperty("message", executionException.getMessage());
+    getHttpErrorCode(executionException)
+        .ifPresent(i -> exceptionEntry.addProperty("errorCode", String.valueOf(i)));
+
+    return exceptionEntry;
+  }
+
+  /** How to execute this operation */
```
gpolumbo-broad, line 130: Add `@FunctionalInterface` to this interface definition and to `CowSerialize`

yonghaoy: done

PR #7, opened 2020-06-01
CA-749 Add initial StorageCow and BucketCow

wchamber:
Implement create/get/delete for bucket to StorageCow and delete to BucketCow.

(The diff is messy at the moment because I included `yyu-CA-806` branch and master. Ideally, the logging branch would get in soon and this would look cleaner, or I can rework my PR to just go from master and not incorprate the logging changes. The relevant changes are in the new storage package + IntegrationCredentials + build.gradle)


src/main/java/bio/terra/cloudres/google/storage/BucketCow.java
```diff
@@ -0,0 +1,48 @@
...
+          @Override
+          public Boolean execute() {
+            return bucket.delete();
+          }
+
+          @Override
+          public String serializeRequest() {
+            return bucket.getName();
+          }
+        });
```
yonghaoy, line 46: This class is gone and has changed to to 3 arguments

src/main/java/bio/terra/cloudres/google/storage/BucketCow.java
```diff
@@ -0,0 +1,48 @@
...
+          }
+
+          @Override
+          public Boolean execute() {
+            return bucket.delete();
+          }
+
+          @Override
+          public String serializeRequest() {
+            return bucket.getName();
```
yonghaoy, line 44: why we just want the bucket.name()?

src/main/java/bio/terra/cloudres/google/storage/StorageCow.java
```diff
@@ -0,0 +1,96 @@
...
+
+              @Override
+              public Bucket execute() {
+                return storage.create(bucketInfo);
+              }
+
+              @Override
+              public String serializeRequest() {
+                Gson gson = new Gson();
+                return gson.toJson(bucketInfo, BucketInfo.class);
```
yonghaoy, line 47: inline those two: new Gosn().toJson(bucketInfo, BucketInfo.class);

src/main/java/bio/terra/cloudres/google/storage/StorageCow.java
```diff
@@ -0,0 +1,96 @@
...
+        operationAnnotator.executeCowOperation(
+            new CowOperation<Bucket>() {
+              @Override
+              public CloudOperation getCloudOperation() {
+                return CloudOperation.GOOGLE_GET_BUCKET;
+              }
+
+              @Override
+              public Bucket execute() {
+                return storage.get(bucket);
```
yonghaoy, line 65: I see we can pass BucketGetOption:
https://github.com/googleapis/java-storage/blob/a3b04ddd528fe066a750ae4d0d5626ca43759b32/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java#L1935

Do we also want this?

yonghaoy: The same question for delete.

src/test/java/bio/terra/cloudres/google/storage/IntegrationUtils.java
```diff
@@ -0,0 +1,33 @@
...
+import com.google.auth.oauth2.ServiceAccountCredentials;
+import com.google.cloud.storage.StorageOptions;
+import java.util.UUID;
+
+/** Utilities for integration tests of the storage package. */
+class IntegrationUtils {
+  private IntegrationUtils() {}
+
+  static final ClientConfig DEFAULT_CLIENT_CONFIG =
+      ClientConfig.Builder.newBuilder().setClient("integration").build();
```
yonghaoy, line 14: make "integration" a constant? And shall we have their own constant or you are expecting each package will have their own client name?

src/test/java/bio/terra/cloudres/google/storage/IntegrationUtils.java
```diff
@@ -0,0 +1,33 @@
+package bio.terra.cloudres.google.storage;
+
+import bio.terra.cloudres.common.ClientConfig;
+import bio.terra.cloudres.testing.IntegrationCredentials;
+import com.google.auth.oauth2.ServiceAccountCredentials;
+import com.google.cloud.storage.StorageOptions;
+import java.util.UUID;
+
+/** Utilities for integration tests of the storage package. */
+class IntegrationUtils {
```
yonghaoy, line 10: I feel this should be in testing/ dir, not in the test/.

src/test/java/bio/terra/cloudres/google/storage/BucketCowTest.java
```diff
@@ -0,0 +1,19 @@
...
+
+@Tag("integration")
+public class BucketCowTest {
+  @Test
+  public void deleteCreatedBucket() {
+    StorageCow storageCow = IntegrationUtils.defaultStorageCow();
+    String bucketName = IntegrationUtils.randomName();
+    BucketCow bucketCow = storageCow.create(BucketInfo.of(bucketName));
+
+    assertTrue(bucketCow.delete());
```
yonghaoy, line 17: do we want to test storageCow.create() acctually creates things, and delete actually delete the cloud object?  like
BucketCow bucketCow = storageCow.create(BucketInfo.of(bucketName));
assertBucketExists(bucketName);
assertTrue(bucketCow.delete());
assertBucketNotExists(bucketName);

src/main/java/bio/terra/cloudres/google/storage/StorageCow.java
wchamber: Good question. I omitted the array arguments that weren't used by DataRepo since they aren't used. I could see us starting by supporting them, or waiting until they're needed by clients. Wasn't sure what was better, opted for starting with less work. WDYT?

src/test/java/bio/terra/cloudres/google/storage/BucketCowTest.java
wchamber: added two StorageCow#gets. It's a little self-refrential, that you have to trust the StorageCow actually is getting something, but it's calling the underlying Google library stuff.

src/test/java/bio/terra/cloudres/google/storage/IntegrationUtils.java
wchamber: Hmm, I moved some of it to the common testing directory, but left the Storage specific stuff in a file in this directory. I don't want to have cloud package specific stuff (like StorageCow and StorageOptions) in the common testing directory. WDYT?

src/test/java/bio/terra/cloudres/google/storage/IntegrationUtils.java
wchamber: Good point, made it a constant in the other file, see below comment.

src/main/java/bio/terra/cloudres/google/storage/BucketCow.java
wchamber: Hmm, we want to identify the bucket for the delete. Maybe the project id + the bucket name? I'm thinking about making a storage/SerializationUtils, which would include a`JsonObject convertBucketandProject(String bucketName, string projectId)`, and use that here and in the get/delete bucket in StorageCow. WDYT?

yonghaoy: I think what we want to log is whatever the request looks like(or something enough for us seeking help for google when filing gcp support). If that is just bucketName, that is enough. 
Also, in the near future, as discussed in anouther comment, we may also want to log BucketOptions if present. I am fine without projectId for now since they will change eventually soon?

src/main/java/bio/terra/cloudres/google/storage/StorageCow.java
yonghaoy: SG!

src/test/java/bio/terra/cloudres/google/storage/IntegrationUtils.java
yonghaoy: How about creates testing/ dir under main/...storage/testing dir or javatests/...storage/testing, if that is storage specific? My understand of file structure is that: files under /test should be tests for the code corresponding to source code in main/.
To difficientate testUtils vs actually tests, the utils should be in speperate directory?

src/test/java/bio/terra/cloudres/google/storage/BucketCowTest.java
yonghaoy: right, we are using the code StorageCow when testing StorageCow..But since you are writting integration tests, we must to have a way to actually get the bucket after creating/deleting it. .But I don't have a better option as well. And I feel that would be the common problem for all cows.
Another related question: shall we have unit test + integration? The RMCow only have unit tests(that has less value I feel) and this only have integration test(also has this issue). For other future cows, which patten should it follow? Just integration?

src/test/java/bio/terra/cloudres/google/storage/IntegrationUtils.java
wchamber: I don't think that's a hard rule. Internal Google style is to put testing utils shared across packages under java/, but explicitly allows within-package testing utils to live in the javatests/ package where they're used. In Google, they have the luxury of `testonly` build commands, so I think it's reasonable for us to put testing only utils shared across packages in javatests. So I think it's ok, but we should be consistent. go/java-style#s8.2.1-java-and-javatests-files

src/main/java/bio/terra/cloudres/google/storage/BucketCow.java
wchamber: I think it's just the bucket-name - not even the project id is needed. bucket names are in a single global namespace. https://cloud.google.com/storage/docs/naming-buckets#considerations

src/test/java/bio/terra/cloudres/google/storage/BucketCowTest.java
wchamber: I think the StorageCow gets are good enough.
I don't think there's anything to unittest about the storage cow itself. I think we might want to unittest serialization of requests, but so far that's about it. I don't think it's worth writing typo tests to verify the right CloudOperation is passed to every OperationAnnotator call.

wchamber: I think 1). Until it's used by other files, just keep it in the file with @VisibleForTesting. Then add a unittest for the Cow where the unittests are only for that convert.

PR #8, opened 2020-06-03
CA-829 Add BlobCow and blob operations

wchamber:
Add Blob operations and some serialization utiltiies.
Add integration tests for blobs. Introduce buckets to be reused for a test class in an integration test.


src/main/java/bio/terra/cloudres/google/storage/BlobCow.java
```diff
@@ -0,0 +1,60 @@
...
+    return blob;
+  }
+
+  /** See {@link Blob#copyTo(BlobId, Blob.BlobSourceOption...)} */
+  public CopyWriter copyTo(BlobId targetblob) {
+    return operationAnnotator.executeCowOperation(
+        CloudOperation.GOOGLE_COPY_BLOB,
+        () -> blob.copyTo(targetblob),
+        () -> {
+          // TODO probably need to adjust for serialization.
```
yonghaoy: nit, not sure what is the TODO format would looks like, I just keep using TODO(yonghao):

src/test/java/bio/terra/cloudres/google/storage/BlobCowTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+    BlobCow blob = storageCow.create(BlobInfo.newBuilder(blobId).build());
+    assertEquals(storageCow.get(blobId).getBlobInfo().getName(), blobId.getName());
+
+    assertTrue(blob.delete());
+    assertNull(storageCow.get(blobId));
+  }
+
+  @Test
+  public void copyTo() throws IOException {
```
yonghaoy, line 62: nit: You can just let it throw raw exception:
http://go/java-practices/exceptions#test_methods

src/test/java/bio/terra/cloudres/google/storage/StorageCowTest.java
```diff
@@ -3,12 +3,41 @@
...
+
+  @BeforeAll
+  public static void createReusableBucket() {
+    reusableBucket =
+        StorageIntegrationUtils.defaultStorageCow()
+            .create(BucketInfo.of(IntegrationUtils.randomName()));
+  }
+
+  @AfterAll
+  public static void deleteReusableBucket() {
```
yonghaoy, line 35: I usually name it as tearDownTest and setUpTest just in case we need to do multiple things. But it is up to you., didn't find any best practice reference

src/test/java/bio/terra/cloudres/google/storage/StorageCowTest.java
```diff
@@ -23,4 +52,36 @@ public void createGetDeleteBucket() {
...
+    BlobId blobId =
+        BlobId.of(reusableBucket.getBucketInfo().getName(), IntegrationUtils.randomName());
+    assertNull(storageCow.get(blobId));
+
+    String contents = "hello blob";
+    try (WriteChannel writeChannel = storageCow.writer(BlobInfo.newBuilder(blobId).build())) {
+      writeChannel.write(ByteBuffer.wrap(contents.getBytes(StandardCharsets.UTF_8)));
+    }
+    BlobCow blob = storageCow.get(blobId);
+    assertEquals(BlobCowTest.readContents(blob), contents);
```
yonghaoy, line 84: nit, the order should be assertEquals(expected, actual).  Same for others

src/test/java/bio/terra/cloudres/google/storage/BlobCowTest.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  @Test
+  public void reader() throws IOException {
+    BlobId blobId =
+        BlobId.of(reusableBucket.getBucketInfo().getName(), IntegrationUtils.randomName());
+
+    final String contents = "hello my blob";
+    BlobCow blob = createBlobWithContents(blobId, contents);
+    assertEquals(readContents(blob), contents);
```
yonghaoy, line 89: nit new line between 88,89?

src/main/java/bio/terra/cloudres/google/storage/BlobCow.java
wchamber: yeah. I think if we have a ticket, TODO'ing the ticket number is better. for this, it was an older TODO and I removed it.

src/test/java/bio/terra/cloudres/google/storage/StorageCowTest.java
```diff
@@ -3,12 +3,41 @@
...
+  @AfterAll
+  public static void deleteReusableBucket() {
+    // This only succeeds if the reusableBucket is empty. If it's not empty, something about the
+    // tests has failed.
+    reusableBucket.delete();
+  }
+
   @Test
   public void createGetDeleteBucket() {
     StorageCow storageCow = StorageIntegrationUtils.defaultStorageCow();
```
yonghaoy, line 43: do we need to create cow in each method?

src/test/java/bio/terra/cloudres/google/storage/BlobCowTest.java
wchamber: thanks, done.

src/test/java/bio/terra/cloudres/google/storage/StorageCowTest.java
wchamber: done

src/test/java/bio/terra/cloudres/google/storage/BlobCowTest.java
wchamber: done

src/test/java/bio/terra/cloudres/google/storage/StorageCowTest.java
wchamber: nope. moved it up to the class.

src/main/java/bio/terra/cloudres/google/storage/StorageCow.java
```diff
@@ -26,16 +27,39 @@ public StorageCow(ClientConfig clientConfig, StorageOptions storageOptions) {
...
+  public BlobCow get(BlobId blob) {
+    Blob rawBlob =
+        operationAnnotator.executeCowOperation(
+            CloudOperation.GOOGLE_GET_BLOB,
+            () -> storage.get(blob),
+            () -> SerializeUtils.convert(blob));
+    if (rawBlob == null) {
+      return null;
+    }
+    return new BlobCow(clientConfig, rawBlob);
```
yonghaoy, line 60: optional: return rawBlob == null ? null : new BlobCow(clientConfig, rawBlob)

src/main/java/bio/terra/cloudres/google/storage/StorageCow.java
```diff
@@ -26,16 +27,39 @@ public StorageCow(ClientConfig clientConfig, StorageOptions storageOptions) {
...
     Bucket bucket =
         operationAnnotator.executeCowOperation(
             CloudOperation.GOOGLE_CREATE_BUCKET,
             () -> storage.create(bucketInfo),
-            () -> new Gson().toJsonTree(bucketInfo, BucketInfo.class).getAsJsonObject());
+            () -> SerializeUtils.convert(bucketInfo));
     return new BucketCow(clientConfig, bucket);
   }
 
+  /** See {@link Storage#get(BlobId)}. Returns null if blob is not found. */
```
yonghaoy, line 50: worth to add a javadoc @return saying it may return null?

src/main/java/bio/terra/cloudres/google/storage/BlobCow.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  /** See {@link Blob#copyTo(BlobId, Blob.BlobSourceOption...)} */
+  public CopyWriter copyTo(BlobId targetblob) {
+    return operationAnnotator.executeCowOperation(
+        CloudOperation.GOOGLE_COPY_BLOB,
+        () -> blob.copyTo(targetblob),
+        () -> {
+          JsonObject request = new JsonObject();
+          request.add("source", SerializeUtils.convert(blob.getBlobId()));
```
yonghaoy, line 38: What if we add json tree here, instead of convert to JsonTree then JsonObject?(then this is simple enough without the SerializeUtils)
Gson gson = new Gson();
request.add("source", gson.toJsonTree(blob.getBlobId())));
request.add("target", gson.toJsonTree(targetblob));

will that work?

src/test/java/bio/terra/cloudres/google/storage/StorageCowTest.java
yonghaoy: sorry, one more point I just find when writting test. In my case, I want to call setup and tearDown using beforeEach.  Because one test updates the ACL and I want to reset in next test.  In this case, it is fine to reset the resource for each test method actually. 

Also, if you change to BeforeEach, no static method are required, you can just create a storageCow in the class level, and get rid of the StorageIntegrationUtils.defaultStorageCow method?

wchamber: That test sounds like it does care about the state of the thing being setUp/tornDown. In that case, using a single static BeforeAll/AfterAll doesn't sound right- you want the thing to be in a known state at the start of each test, or make a thing specifically for that test and make it have the state once.

Not sure what you mean, there is a storageCow at the class level for this now. The method for creatng the storage cow is used in different files.

src/main/java/bio/terra/cloudres/google/storage/StorageCow.java
```diff
@@ -26,16 +27,39 @@ public StorageCow(ClientConfig clientConfig, StorageOptions storageOptions) {
     this.storage = storageOptions.getService();
   }
 
+  /** See {@link Storage#create(BlobInfo, Storage.BlobTargetOption...)}. */
+  public BlobCow create(BlobInfo blobInfo) {
```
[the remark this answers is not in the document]
wchamber: Mmm, I'll push back a little. that's kind of like having a TODO for every API that we haven't done yet, it's a little implicit. It's a lot of TODOs without a lot of information content. wdyt?

src/main/java/bio/terra/cloudres/google/storage/StorageCow.java
wchamber: think it's overkill and is clear enough in the main sentence to not need explicit calling out. go/java-practices/javadoc#return

src/main/java/bio/terra/cloudres/google/storage/BlobCow.java
wchamber: I still want to serialize the blobId in a consistent way and test that it's serialized correctly, so the SerializeUtils is a feature not bug I think here.

src/main/java/bio/terra/cloudres/google/storage/StorageCow.java
wchamber: done

src/test/java/bio/terra/cloudres/google/storage/StorageCowTest.java
```diff
@@ -3,24 +3,81 @@
...
+    BlobId blobId =
+        BlobId.of(reusableBucket.getBucketInfo().getName(), IntegrationUtils.randomName());
+    assertNull(storageCow.get(blobId));
+
+    String contents = "hello blob";
+    try (WriteChannel writeChannel = storageCow.writer(BlobInfo.newBuilder(blobId).build())) {
+      writeChannel.write(ByteBuffer.wrap(contents.getBytes(StandardCharsets.UTF_8)));
+    }
+    BlobCow blob = storageCow.get(blobId);
+    assertEquals(contents, BlobCowTest.readContents(blob));
```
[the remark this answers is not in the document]
wchamber: done

src/main/java/bio/terra/cloudres/google/storage/StorageCow.java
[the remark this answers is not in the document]
yonghaoy: okay with that as well. If we won't forget that we can leave it as it.  But seems that's not lot work, worth to just add those Options for other cows we implement. I don't think we need a follow up PR just for the options in the future. But for this, we can leave it empty.

src/main/java/bio/terra/cloudres/google/storage/StorageCow.java
yonghaoy: okay. as long as there is some place we can indicate this will return null.

src/main/java/bio/terra/cloudres/google/storage/BlobCow.java
yonghaoy: Not sure if that is your case, but in my case, I only need:
gson.toJsonTree(blob.getBlobId()),  and it won't be used in other classes. 
so I feel there is no need to introduce the extra SerializeUtils. 
And for the test, it can just be a unit test to test the convert() method in cowTest? 
That is just my opinion, it doesn't matter feel free to resolve it.

PR #11, opened 2020-06-04
[CA-830] Add BigQueryCow to support update/create/delete/get Bigquery

yonghaoy:
https://broadworkbench.atlassian.net/browse/CA-830


build.gradle
```diff
@@ -44,8 +44,9 @@ dependencies {
     compile 'io.opencensus:opencensus-api:0.26.0'
     runtime 'io.opencensus:opencensus-impl:0.26.0'
 
-    compile 'com.google.cloud:google-cloud-resourcemanager:0.117.2-alpha'
-    compile 'com.google.cloud:google-cloud-storage:1.108.0'
+    compile 'com.google.cloud:google-cloud-resourcemanager:0.117.2-alpha' // Google Resource Manager API Client
+    compile 'com.google.cloud:google-cloud-bigquery:1.116.0' // Google BigQuery API Client
+    compile 'com.google.cloud:google-cloud-storage:1.108.0' // Google Storage API Client
```
wchamber, line 49: nit: remove the comments. What's being added is kind of obvious from the path name.

src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
```diff
@@ -0,0 +1,115 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A Cloud Object Wrapper(COW) for Google API Client Library: {@link ResourceManager}
+ *
+ * <p>Eventually there might be multiple COW classes for each resource type, e.g. ProjectCow.
```
wchamber, line 22: bad copy paste, fix plz

src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
```diff
@@ -0,0 +1,115 @@
...
+  private final OperationAnnotator operationAnnotator;
+  private final BigQuery bigQuery;
+
+  public BigQueryCow(ClientConfig clientConfig, BigQueryOptions bigQueryOptions) {
+    this.operationAnnotator = new OperationAnnotator(clientConfig, logger);
+    this.bigQuery = bigQueryOptions.getService();
+  }
+
+  @VisibleForTesting
+  static JsonObject convert(DatasetInfo datasetInfo, DatasetOption... options) {
```
wchamber, line 36: nit: consider moving all these convert functions to the end of the class. Not exactly the most important thing to start reading when you start with the file.

src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
```diff
@@ -0,0 +1,115 @@
...
+    jsonObject.addProperty("datasetDeleteOptions", new Gson().toJson(options));
+    return jsonObject;
+  }
+
+  /**
+   * Creates {@link Dataset}
+   *
+   * @param datasetInfo The {@link DatasetInfo} to get
+   * @param datasetOptions The array of {@link DatasetOption}
+   * @return the project being created
```
wchamber, line 65: I think @linking the underlying BigQuery method is better than reproducing what it does here. This can be incomplete or misleading, and we're just trying to replicate documentation that already exists.
ditto others.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -0,0 +1,110 @@
...
+ * Note that Dataset has quota about "5 operations every 10 seconds per dataset". So in some update,
+ * delete test, need to create a new dataset everytime.
+ */
+public class BigQueryCowTest {
+  private static final ServiceAccountCredentials GOOGLE_CREDENTIALS =
+      IntegrationCredentials.getGoogleCredentialsOrDie();
+
+  // Dataset id only allows underscore
+  private static final String DATASET_ID = IntegrationUtils.randomName().replace('-', '_');
+  private static final String REUSABLE_DATASET_ID = IntegrationUtils.randomName().replace('-', '_');
```
wchamber, line 23: maybe add some more functions to IntegrationUtils to generate names that match different constraints. This isn't going to be the last naming constraint.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -0,0 +1,110 @@
...
+  private static final DatasetOption DATASET_OPTION_ACCESS =
+      DatasetOption.fields(BigQuery.DatasetField.ACCESS);
+  private static final DatasetOption DATASET_OPTION_CREATION_TIME =
+      DatasetOption.fields(BigQuery.DatasetField.CREATION_TIME);
+  private static final DatasetInfo.Builder REUSABLE_DATASET_INFO_BUILDER =
+      DatasetInfo.newBuilder(REUSABLE_DATASET_ID);
+  private static final DatasetInfo.Builder DATASET_INFO_BUILDER =
+      DatasetInfo.newBuilder(DATASET_ID);
+
+  private static BigQueryOptions defaultStorageOptions() {
```
wchamber, line 34: rename

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -0,0 +1,110 @@
...
+  @Test
+  public void convertDatasetInfoWithOptions() {
+    assertEquals(
+        "{\"datasetId\":{\"datasetId\":{\"dataset\":\""
+            + DATASET_ID
+            + "\"},\"labels\":{\"userMap\":{}}},\"datasetOptions\":[{\"rpcOption\":\"FIELDS\",\"value\":\"datasetReference,access\"},{\"rpcOption\":\"FIELDS\",\"value\":\"datasetReference,creationTime\"}]}",
+        BigQueryCow.convert(
+                DATASET_INFO_BUILDER.build(), DATASET_OPTION_ACCESS, DATASET_OPTION_CREATION_TIME)
+            .toString());
+  }
```
wchamber, line 77: also test the convert dataset delete?
I also wouldn't mind having theese at the end of the file.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+    assertEquals(createdDataSet, bigQueryCow.getDataSet(DATASET_ID));
+    assertEquals(DATASET_ID, createdDataSet.getDatasetId().getDataset());
+  }
+
+  @Test
+  public void getDataset() {
+    assertEquals(
+        REUSABLE_DATASET_ID,
+        bigQueryCow.getDataSet(REUSABLE_DATASET_ID).getDatasetId().getDataset());
```
wchamber, line 91: consider making a single create/get/delete test rather than messing with a shared DATASET_ID and relying on the order the test is run. @Test functions should not care about they're ordering.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  @Test
+  public void updateDataset() {
+    assertNull(bigQueryCow.getDataSet(REUSABLE_DATASET_ID).getDescription());
+    String description = "new description";
+    bigQueryCow.updateDataset(
+        REUSABLE_DATASET_INFO_BUILDER.setDescription("new description").build());
+    assertEquals(description, bigQueryCow.getDataSet(REUSABLE_DATASET_ID).getDescription());
```
wchamber, line 100: why not just create/update/delete a dataset here rather than creating a reusable dataset? It's funny to modify properties on the shared, re-used dataset, and it's also not really getting re-used.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -0,0 +1,110 @@
...
+  private static final String REUSABLE_DATASET_ID = IntegrationUtils.randomName().replace('-', '_');
+
+  private static final DatasetOption DATASET_OPTION_ACCESS =
+      DatasetOption.fields(BigQuery.DatasetField.ACCESS);
+  private static final DatasetOption DATASET_OPTION_CREATION_TIME =
+      DatasetOption.fields(BigQuery.DatasetField.CREATION_TIME);
+  private static final DatasetInfo.Builder REUSABLE_DATASET_INFO_BUILDER =
+      DatasetInfo.newBuilder(REUSABLE_DATASET_ID);
+  private static final DatasetInfo.Builder DATASET_INFO_BUILDER =
+      DatasetInfo.newBuilder(DATASET_ID);
```
wchamber, line 32: Just inline these. There constructions are simple enough that we're not saving a lot of boilerplate, nor do these need to be the same across multiple tests. These are simple, so we can just inline them. go/unit-test-practices#constants

src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
```diff
@@ -0,0 +1,115 @@
...
+   */
+  public Dataset createDataset(DatasetInfo datasetInfo, DatasetOption... datasetOptions) {
```
wchamber, line 67: optional - make this return a DatasetCow, where the DatasetCow for now only has a getDataSetInfo function. We're going to want to return a DatasetCow here eventually. Up to you when you want to do that.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
yonghaoy: the intellij's google-java-format pulgin auto move those methods to top. 
Moved them to bottom, and just use gradlew spotlessapply, not't the intellij's formatter

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
yonghaoy: are you suggesting to have a single test to have them all?
The current test doesn't rely on any order. 
The get one reads the reusable data set, not DATASET_ID.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
yonghaoy: done.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
yonghaoy: done

src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
yonghaoy: I don't quite understand the logic and cow structure now.
I thought each resource type having its own cow would be future thing and we are not sure yet?  For now, things are clear because this wrapped BigQuery and clients knows to use this Cow to delete/create dataset. If we have datasetCow, will deleteDataset goes here or DatasetCow?

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
yonghaoy: wrong reply. not this comment

src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
yonghaoy: the intellij's google-java-format pulgin auto move those methods to top.
Moved them to bottom, and just use gradlew spotlessapply, not't the intellij's form

yonghaoy: here

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
wchamber: Sorry, wasn't very clear. I'm suggesting having one test for create/get/delete that uses its own dataest and one test for update that uses its own dataset. At the very least, let's stop re-using dataset ids.

create/delete both use the same DATASET_ID, but why? Using the same DATASET_ID implies that these tests care about which dataset they're using and it should have the same id. The create test doesn't delete the dataset it creates, but I missed that the delete test also creates the dataset.
If we stop re-using dataset ids between tests, then we know that the ordering doesn't matter and there isn't cross test sharing of datasets. We only want to re-use a resource when we don't actually care about that resource. In the blob tests, it's ok to re-use a bucket since we're just using it as a container. all blobs need to go in a bucket, but tests about those blobs don't care which bucket they're in.

src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
wchamber: Each client cloud class has its own Cow. Each cow wraps the underlying cloud classes API. The google cloud client classes have duplicate functions, e.g. BigQuery and Dataset both have deletion methods. So both the BigQueryCow and the DatasetCow are allowed to have the corresponding deletion methods. Both of those deletes should be the same CloudOperation, but it's ok if there are two.

We don't want BigQueryCow to return a raw Dataset to the client because that encourages them to do things like call delete on that Dataset, which won't be logged by CRL. If we return a DatasetCow instead from BigQueryCow, now the client is encouraged by the CRL API to use Cows correctly.

We're not sure about having Cows for data-plane operations (we don't need to log each call to get the next chunk or page of results of a read), but we know we want them for control-plane, e.g. creating/deleting resources or that a read is being initiated. So TableQuery may not have a Cow since that just has the data results of a query, but there must be a DatasetCow because Dataset has methods like delete. Happy to talk more.
https://docs.google.com/document/d/1C-NL48JHNwhbwH8jHNaoiKdKJDEXBBr4Y3cjAwgeNsA/edit#heading=h.xxvl37disaa0

src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
```diff
@@ -32,84 +27,60 @@ public BigQueryCow(ClientConfig clientConfig, BigQueryOptions bigQueryOptions) {
...
 
-  /**
-   * Deletes {@link Dataset}
-   *
-   * @param datasetId The DatasetId to get
-   * @param deleteOptions The array of {@link DatasetDeleteOption}
-   * @return the project being created
-   */
+  /** See {@link BigQuery#delete(String, DatasetDeleteOption...)}. */
   public Boolean deleteDataset(String datasetId, DatasetDeleteOption... deleteOptions) {
```
wchamber, line 47: nit: use the primitive `boolean` instead of the object `Boolean`

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -19,33 +20,29 @@
...
   private static final DatasetOption DATASET_OPTION_ACCESS =
       DatasetOption.fields(BigQuery.DatasetField.ACCESS);
   private static final DatasetOption DATASET_OPTION_CREATION_TIME =
       DatasetOption.fields(BigQuery.DatasetField.CREATION_TIME);
-  private static final DatasetInfo.Builder REUSABLE_DATASET_INFO_BUILDER =
-      DatasetInfo.newBuilder(REUSABLE_DATASET_ID);
-  private static final DatasetInfo.Builder DATASET_INFO_BUILDER =
-      DatasetInfo.newBuilder(DATASET_ID);
+  private static final DatasetDeleteOption DATASET_OPTION_DELETE_CREATION_TIME =
+      DatasetDeleteOption.deleteContents();
```
wchamber: inline the rest of these that are simple to construct.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -72,39 +102,26 @@ public void convertDatasetInfoWithOptions() {
...
 
   @Test
-  public void updateDataset() {
-    assertNull(bigQueryCow.getDataSet(REUSABLE_DATASET_ID).getDescription());
-    String description = "new description";
-    bigQueryCow.updateDataset(
-        REUSABLE_DATASET_INFO_BUILDER.setDescription("new description").build());
-    assertEquals(description, bigQueryCow.getDataSet(REUSABLE_DATASET_ID).getDescription());
-  }
+  public void createDataset() {
```
wchamber, line 121: group this with the rest of the tests involving datasets before the conversion tests.

src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
yonghaoy: Thanks for explanation.So the rule is to lookup the class, if there are control-plan, some method, we should have a cow for that. 
That is a DFS algorithm, So in this case:
Tow level, we start with BigQueryCow, and then we find Datset, 

https://googleapis.dev/java/google-cloud-clients/0.111.0-alpha/com/google/cloud/bigquery/Dataset.html
DatasetCow eventually will need to have:
(1) DataSetCow delete()
(2) exists()
(3) TableCow createTable()
(4) BigQueryCow getBigQuery() ?????? wow
(5) DatasetCow reload()
(6) TableCow getTable()

And along with that, we will have TableCow,
https://googleapis.dev/java/google-cloud-clients/0.111.0-alpha/com/google/cloud/bigquery/Table.html

and in TableCow, it will have
JobCow copy(table id)

Then we introduce the jobCow, it will have 
https://googleapis.dev/java/google-cloud-clients/0.111.0-alpha/com/google/cloud/bigquery/Job.html
cancel(), reload()? 
And that seems one round end of the DFS...., then go back to BigQueryCow, to do the another round of DFS.

Wow, that is majic

yonghaoy: And I'd prefer that in a follow up pr once this get merged in

wchamber: That's the plan, we'll see if it works in reality. Note that we don't have to return the same instance of BigQueryCow everytime - it should just be using the same ClientConfig.
Follow up PR sg, note there's a separate ticket for TableCow, somaybe do it and JobCow for that ticket, not this one. https://broadworkbench.atlassian.net/browse/CA-831

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -0,0 +1,129 @@
...
+                DatasetOption.fields(BigQuery.DatasetField.ACCESS),
+                DatasetOption.fields(BigQuery.DatasetField.CREATION_TIME))
+            .toString());
+  }
+
+  @Test
+  public void convertDatasetInfoWithOptions() {
+    assertEquals(
+        "{\"datasetId\":{\"datasetId\":{\"dataset\":\""
+            + DATASET_ID
```
wchamber, line 112: make up your own DATASET_ID, or make it test local.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -0,0 +1,129 @@
...
+import bio.terra.cloudres.testing.IntegrationUtils;
+import com.google.auth.oauth2.ServiceAccountCredentials;
+import com.google.cloud.bigquery.*;
+import com.google.cloud.bigquery.BigQuery.DatasetDeleteOption;
+import com.google.cloud.bigquery.BigQuery.DatasetOption;
+import org.junit.jupiter.api.*;
+
+@Tag("integration")
+/**
+ * Note that Dataset has quota about "5 operations every 10 seconds per dataset". So in some update,
```
wchamber, line 15: optional: link the web page that describes the quota.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -0,0 +1,129 @@
...
+ * delete test, need to create a new dataset everytime.
+ */
+public class BigQueryCowTest {
+  private static final ServiceAccountCredentials GOOGLE_CREDENTIALS =
+      IntegrationCredentials.getGoogleCredentialsOrDie();
+
+  // Dataset id only allows underscore
+  private static final String DATASET_ID = IntegrationUtils.randomNameWithUnderscore();
+  private static final String REUSABLE_DATASET_ID = IntegrationUtils.randomNameWithUnderscore();
+
```
wchamber, line 25: do we still need these (and the reusabledataset static methods) still?

yonghaoy: nope

src/main/java/bio/terra/cloudres/common/CloudOperation.java
```diff
@@ -2,18 +2,24 @@
...
   GOOGLE_COPY_BLOB,
   GOOGLE_CREATE_BLOB,
   GOOGLE_CREATE_BLOB_AND_WRITER,
   GOOGLE_CREATE_BUCKET,
   GOOGLE_GET_BLOB,
   GOOGLE_GET_BUCKET,
-  GOOGLE_DELETE_BLOB,
   GOOGLE_DELETE_BUCKET,
+  GOOGLE_DELETE_BLOB,
   GOOGLE_READ_BLOB;
```
yonghaoy, line 18:
```suggestion
  GOOGLE_READ_BLOB,
```

PR #17, opened 2020-06-10
[CA-831] Support Table create/update/delete/get. Also add a generic Page Impl for all Cows

yonghaoy:
Ticket: https://broadworkbench.atlassian.net/browse/CA-831


src/main/java/bio/terra/cloudres/google/bigquery/TableCow.java
```diff
@@ -0,0 +1,67 @@
...
+            CloudOperation.GOOGLE_UPDATE_BIGQUERY_TABLE,
+            () -> table.update(TableOptions),
+            () -> convert(table, TableOptions)));
+  }
+
+  /** See {@link Table#exists()} */
+  public boolean exists() {
+    return operationAnnotator.executeCowOperation(
+        CloudOperation.GOOGLE_DELETE_BIGQUERY_TABLE,
+        table::exists,
```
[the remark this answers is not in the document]
wchamber: yep, that's right, it's equivalent. It's syntax sugar, so it's very mildly preferred (intellij might suggest the change. as a reviewer, I probably wouldn't notice and it'd always be at the 'nitpick' level. I would do it if I noticed, but I think I've already merged some that I didn't notice :) ).
We can't use `::` on something like the () -> convert(...) because the arguments to convert are not the same as the arguments for the lambda.
here's the oracle doc about these 'Method References' for reference: https://docs.oracle.com/javase/tutorial/java/javaOO/methodreferences.html

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -15,53 +19,160 @@
...
+  }
+
+  @BeforeEach
+  public void setUp() {
+    // Cleanup the tableId and datesetId list and use them to track resources created in one test
+    // method.
+    // In this way to can make sure tables/datasets can always be cleaned up even test failed in the
+    // middle.
+    createdTableIds = new ArrayList<>();
+    createdDatasetIds = new ArrayList<>();
```
wchamber, line 49: you can inline the new ArrayList<>() construction on lines 28-29. Keep the comments on them. The BigQueryCowTest is instantiated for each @Test method.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -15,53 +19,160 @@
...
+  public static void deleteReusableDataset() {
+    bigQueryCow.delete(REUSABLE_DATASET_ID);
+  }
+
+  @BeforeEach
+  public void setUp() {
+    // Cleanup the tableId and datesetId list and use them to track resources created in one test
+    // method.
+    // In this way to can make sure tables/datasets can always be cleaned up even test failed in the
+    // middle.
```
wchamber, line 47: I actually don't care much about cleaning up the resources if the tests failed. It's nice to do on successful tests, but once our test is failing it's impossible to do correctly in all of the cases.

Imagine the tearDown fails on one of the createdTableIds (either because of one of our issues, or a connection issue, or a Google issue, etc.), then not all of them are deleted and the datasets aren't deleted.
So it's fine to do some best effort, but we shouldn't over promise in these comments that they will "always be cleaned up." You've probably already leaked some resources writing these tests: this is why we want the janitor in the first place. I know I have leaked buckets writing my tests.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -15,53 +19,160 @@
...
+    createdDatasetIds.forEach(bigQueryCow::delete);
+  }
 
   @Test
   public void createDataset() {
     String datasetId = IntegrationUtils.randomNameWithUnderscore();
     DatasetInfo createdDataSet =
-        bigQueryCow.createDataset(DatasetInfo.newBuilder(datasetId).build()).getDatasetInfo();
+        bigQueryCow.create(DatasetInfo.newBuilder(datasetId).build()).getDatasetInfo();
+    createdDatasetIds.add(datasetId);
```
wchamber, line 63: there are too many implicit assumptions with this that are not enforced/encouraged.
1. Everything added to createdDatasetIds is from the same BigQueryCow.
2. Whenever a Dataset is created, it's added to createdDatasetIds.

If you want to use the createdDatasetIds for tracking, make and use a helper function to create Datasets, e.g.
```
DatasetCow createTrackedDataset(DatasetInfo datasetInfo) {
  DatasetCow dataset = bigQueryCow.create(datasetInfo);
  createdDatasetIds.add(datasetId);
  return dataset;
}
```
Or leave it as manual tracking, but the automated tracking is probably nicer.
Ditto for tracked tables.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryIntegrationUtils.java
```diff
@@ -17,4 +27,39 @@ static BigQueryCow defaultBigQueryCow() {
...
+    assertTrue(
+        expect.stream()
+            .map(TableId::getDataset)
+            .collect(Collectors.toList())
+            .containsAll(
+                actualTableIds.stream().map(TableId::getDataset).collect(Collectors.toList())));
+  }
+
+  static TableCow createTableCow(
+      BigQueryCow bigQueryCow, String datasetId, List<TableId> createdTableIds) {
```
wchamber, line 57: This feels odd to try to re-use as a static method. It has a lot of assumptions about it's arguments: the datasetId is already created, something is going to happen to delete the createdTableIds. The rest of that logic is being spread across different files.
If we're trying to bundle data and behavior to be re-used, we want a class that encapsulates that. sketch:
```
class TableTracker {
  private final BigQueryCow;
  private final String datasetId; // must be a created dataset comment assumption.
  private final List<TableId> createdTables;

  public TableCow create() {...}
  public void tearDown() {...}  // deletes createdTables, must be done before the dataset is deleted.
}
```
Consider if this is worth doing with datasets as well - it really is a judgment call. This is more complexity, infrastructure, maintenance burden for our tests, but it makes it easier to re-use and write more tests. But once we're trying to re-use behavior, it's better to use classes than smearing the logic across data parameters.

src/test/java/bio/terra/cloudres/google/bigquery/SerializeUtilsTest.java
```diff
@@ -3,31 +3,38 @@
...
 
 @Tag("unit")
 public class SerializeUtilsTest {
+  private final DatasetId datasetId = DatasetId.of("datasetId1");
+  private final String jsonDatasetId = "{\"datasetId\":{\"dataset\":\"datasetId1\"}";
+  private final TableId tableId = TableId.of(datasetId.getDataset(), "tableId1");
+  private final String jsonTableId =
+      "{\"tableId\":{\"dataset\":\"datasetId1\",\"table\":\"tableId1\"}";
+
+  private static final String REUSABLE_DATASET_ID = IntegrationUtils.randomNameWithUnderscore();
```
wchamber, line 18: I think we can just use "123" - we don't need a randomly changing unique name for serialize tests.
And there's nothing being "reused" here.

src/main/java/bio/terra/cloudres/google/bigquery/TableCow.java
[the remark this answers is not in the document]
yonghaoy: yes, it is the same as table::exists, and that is the intellij's suggestion to use lambda if possible.

src/test/java/bio/terra/cloudres/google/bigquery/SerializeUtilsTest.java
```diff
@@ -3,31 +3,38 @@
...
+import com.google.cloud.bigquery.*;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
 @Tag("unit")
 public class SerializeUtilsTest {
+  private final DatasetId datasetId = DatasetId.of("datasetId1");
+  private final String jsonDatasetId = "{\"datasetId\":{\"dataset\":\"datasetId1\"}";
+  private final TableId tableId = TableId.of(datasetId.getDataset(), "tableId1");
+  private final String jsonTableId =
```
wchamber, line 15: i think these should be `private static final`. They're constants, not member variables.

src/test/java/bio/terra/cloudres/google/bigquery/SerializeUtilsTest.java
```diff
@@ -3,31 +3,38 @@
...
   @Test
   public void convertDatasetInfoWithOptions() {
-    String datasetId = IntegrationUtils.randomNameWithUnderscore();
     assertEquals(
-        "{\"datasetId\":{\"datasetId\":{\"dataset\":\""
-            + datasetId
-            + "\"},\"labels\":{\"userMap\":{}}},\"datasetOptions\":[{\"rpcOption\":\"FIELDS\",\"value\":\"datasetReference,access\"},{\"rpcOption\":\"FIELDS\",\"value\":\"datasetReference,creationTime\"}]}",
+        "{\"datasetInfo\":"
+            + jsonDatasetId
+            + ",\"labels\":{\"userMap\":{}}},\"datasetOptions\":[{\"rpcOption\":\"FIELDS\",\"value\":\"datasetReference,access\"},{\"rpcOption\":\"FIELDS\",\"value\":\"datasetReference,creationTime\"}]}",
```
wchamber, line 37: I'm not really sold that the 'jsonDatasetId' is making this test any clearer or easier to read. Ditto for the other `json` reused variables. I think I might rather just see all of the json without any variable re-use. WDYT?

yonghaoy: I don't have strong preference, just feel it saves me lot copy/paste work and makes those test method created json string cleaner and nicer.  I created this variable because most of the test method are using the datasetId variable and it is straight forward to check that the datasetId in json format would be the jsonDatasetId. 
I feel instead of make all tests copy/paste the same piece of string, I'd rather create a variable, and let those tests focus more on the rest of json sting.

src/test/java/bio/terra/cloudres/google/bigquery/SerializeUtilsTest.java
yonghaoy: right, that is not used anymore

src/test/java/bio/terra/cloudres/google/bigquery/SerializeUtilsTest.java
wchamber: My inclination would be that it's easier for the reader to see what's going on with the whole json string there. the datasetid is a part of the input, easier to explicitly see it as a part of the output. I'd rather optimize for the reader than the writer, but I won't hold up the review on it.

src/main/java/bio/terra/cloudres/common/CowPageImpl.java
```diff
@@ -0,0 +1,47 @@
+package bio.terra.cloudres.common;
+
+import static com.google.common.collect.Iterables.transform;
+
+import com.google.api.gax.paging.Page;
+import com.google.common.base.Function;
+
+/** Implementation for {@link Page} interface for Cow */
+public abstract class CowPageImpl<T, C> implements Page<C> {
```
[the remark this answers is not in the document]
yonghaoy: done. Thanks

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
yonghaoy: done

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
yonghaoy: Table Id is already has a helper function, I was too lazy to do it for dataset. 
Now dataset has its own function as well.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
yonghaoy: I had the problem where 
(1) Update test
(2)I didn't generate random id,
(3) First time it failed I compare a wrong field after the update
(4)2nd time it failed because the table id already exists. 
That reminds me that until we have the janitor, we should have some resource cleanup merchinizam (or at lease best effort)
So I choose this way, at least it can do the best effort cleanup(didn't handle the tearDown failed case you mentioned). Because that is the simplest way/best thing I can do. Added "best effort" in comments.

src/test/java/bio/terra/cloudres/google/bigquery/SerializeUtilsTest.java
yonghaoy: done. Some json string is too long to read, and gradle spotless won't give them new lines...

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryIntegrationUtils.java
yonghaoy: (1) This is just function to reduce the copy-pased code when creating dataset/table.
(2) I don't think there would be plan to reuse datasetIds, for reused ones, those would be in @afterAll in each test files. 
(3) This is a temporary solution before janitor cleanup happens. I expected an easier refactor to switch to the new janitor stub with this static helper when I write this. For now, this is just the "best effort"
(4) Table tests are almost done, just two more query test too add
(4) If you think it is worth to do the refactor, I can do a refactor for dataset and tables. Happy to chat more

src/test/java/bio/terra/cloudres/google/bigquery/SerializeUtilsTest.java
```diff
@@ -2,26 +2,19 @@
...
-      "{\"tableId\":{\"dataset\":\"datasetId1\",\"table\":\"tableId1\"}";
-
-  private static final String REUSABLE_DATASET_ID = IntegrationUtils.randomNameWithUnderscore();
 
   @Test
   public void convertDatasetIdWithOptions() {
     assertEquals(
-        jsonDatasetId
-            + ",\"datasetOptions\":[{\"rpcOption\":\"FIELDS\",\"value\":\"datasetReference,access\"},{\"rpcOption\":\"FIELDS\",\"value\":\"datasetReference,creationTime\"}]}",
+        "{\"datasetId\":{\"dataset\":\"datasetId1\"},\"datasetOptions\":[{\"rpcOption\":\"FIELDS\",\"value\":\"datasetReference,access\"},{\"rpcOption\":\"FIELDS\",\"value\":\"datasetReference,creationTime\"}]}",
```
wchamber, line 17: If you manually break up the string with ".." + "..." can this split across lines and not make spotlessApply angry?

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryIntegrationUtils.java
wchamber: 1) Yes, reducing copy-pasted code is good, I'lll argue strongly just having this static method isn't enough.
2) Sorry, wasn't trying to say the *datasetIds* should be re-used, what is being copy/pasted is the rest of the cleanup behavior - initializing the array lists and iterating through them to tear them down.
3) With the janitor, there doesn't have to be any code in these tests to do cleanups, so there's not a refactor here. The tracking will happen in the non-test main CRL code.
I'm split on whether it's worth doing best effor clean-up here at all. On the one hand, we don't want to leave around a bunch of garbage. On the other hand, we can just delete/recreate the integration test project to get the container. So if we want to say we don't want to do best-effort cleanup in these tests, I kind of support that.
4) If we're going to do best effort  clean up, even for a little while, than I think this should be refactored to add best effort Tracker classes. Test code is real code.

src/test/java/bio/terra/cloudres/google/bigquery/SerializeUtilsTest.java
```diff
@@ -2,26 +2,19 @@
...
-import bio.terra.cloudres.testing.IntegrationUtils;
 import com.google.cloud.bigquery.*;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
 @Tag("unit")
 public class SerializeUtilsTest {
   private final DatasetId datasetId = DatasetId.of("datasetId1");
-  private final String jsonDatasetId = "{\"datasetId\":{\"dataset\":\"datasetId1\"}";
   private final TableId tableId = TableId.of(datasetId.getDataset(), "tableId1");
```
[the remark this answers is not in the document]
yonghaoy: Change to static final.

src/main/java/bio/terra/cloudres/common/CloudOperation.java
```diff
@@ -25,5 +25,11 @@
   GOOGLE_UPDATE_DATASET,
   GOOGLE_DELETE_DATASET,
   GOOGLE_GET_DATASET,
-  GOOGLE_RELOAD_DATASET;
+  GOOGLE_RELOAD_DATASET,
```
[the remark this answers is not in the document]
yonghaoy: sorry, still not used to github codereview. Just saw this comments. Ordered now.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryResourceTracker.java
```diff
@@ -0,0 +1,50 @@
...
+import com.google.cloud.bigquery.DatasetInfo;
+import com.google.cloud.bigquery.StandardTableDefinition;
+import com.google.cloud.bigquery.TableId;
+import com.google.cloud.bigquery.TableInfo;
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * Helper class to track BigQuery resources created in tests, this used to clean up resources with
```
wchamber, line 13: english nit: '... created in tests and do best effort clean up.'

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryResourceTracker.java
```diff
@@ -0,0 +1,50 @@
...
+ * best effort.
+ */
+public class BigQueryResourceTracker {
+  private final BigQueryCow bigQueryCow;
+  private final String datasetId; // must be a created dataset comment assumption.
```
wchamber, line 18: nit: /* javadoc */ style comment for member variables and maybe comment what the assumption is about the datasetId instead of "comment assumption"

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryResourceTracker.java
```diff
@@ -0,0 +1,50 @@
...
+  private final List<TableId> createdTableIds;
+  private final List<String> createdDatasetIds;
+
+  public BigQueryResourceTracker(BigQueryCow bigQueryCow, @Nullable String datasetId) {
```
wchamber, line 22: what does it mean for the datasetId to be null?

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryResourceTracker.java
```diff
@@ -0,0 +1,50 @@
...
+import com.google.cloud.bigquery.TableInfo;
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * Helper class to track BigQuery resources created in tests, this used to clean up resources with
+ * best effort.
+ */
+public class BigQueryResourceTracker {
```
wchamber, line 16: optional: consider if this can just be 'ResourceTracker' as everything in this package is about BigQuery.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryResourceTracker.java
```diff
@@ -0,0 +1,50 @@
...
+  private final BigQueryCow bigQueryCow;
+  private final String datasetId; // must be a created dataset comment assumption.
+  private final List<TableId> createdTableIds;
+  private final List<String> createdDatasetIds;
+
+  public BigQueryResourceTracker(BigQueryCow bigQueryCow, @Nullable String datasetId) {
+    this.bigQueryCow = bigQueryCow;
+    this.datasetId = datasetId;
+    createdTableIds = new ArrayList<>();
+    createdDatasetIds = new ArrayList<>();
```
wchamber, line 26: nit: inline these initialization on lines 19-20.

src/test/java/bio/terra/cloudres/google/bigquery/DatasetCowTest.java
```diff
@@ -5,37 +5,30 @@
 
 import bio.terra.cloudres.testing.IntegrationUtils;
 import com.google.cloud.bigquery.*;
-import java.util.ArrayList;
-import java.util.List;
 import org.junit.jupiter.api.*;
 
 @Tag("integration")
 public class DatasetCowTest {
   private static final String REUSABLE_DATASET_ID = IntegrationUtils.randomNameWithUnderscore();
```
wchamber, line 12: WDYT about incorporating the RESUABLE_DATASET_ID in the BigQueryResourceTracker? Instead of having a member BigQueryResourceTracker for each test, we could have a static one that creates this re-used dataset id, tracks all datasets & tables and deletes all of them (including the resuable dataset) in @AfterAll.

src/test/java/bio/terra/cloudres/google/bigquery/BigQueryResourceTracker.java
yonghaoy: As we talked, this can just be used for tracking dataset resources, e.g, in one test:
`BigQueryResourceTracker tracker = new BigQueryResourceTracker(bigqueryCow, null);
`
`tracker.createDatasetCow();`
`tracker.tearDown();`

added checkNotNull section in createdTableCow method(); 
Does it lg?

src/test/java/bio/terra/cloudres/google/bigquery/DatasetCowTest.java
yonghaoy: (1)I'd prefer cleanup the tables after each test, and make the dataset clean and ready for next test. So I didn't use @afterall.
So if we put RESUABLE_DATASET_ID into, the tracker would become complex because we need to have tearDownTable(), tearDownDataset(). 
(2) And if we want the resuable_dataset_id in tracker, I feel that gives us less flexibility if we want to create table in another dataset.
(3) The purpose of the class is a resource tracker and should be totally independent. So I am concern to bring such a restrict into. 
Happy to chat more. 

Thanks

google-bigquery/src/test/java/bio/terra/cloudres/google/bigquery/ResourceTracker.java
```diff
@@ -9,24 +11,21 @@
...
- */
-public class BigQueryResourceTracker {
+/** Helper class to track BigQuery resources created in tests and do best effort clean up. */
+public class ResourceTracker {
   private final BigQueryCow bigQueryCow;
-  private final String datasetId; // must be a created dataset comment assumption.
-  private final List<TableId> createdTableIds;
-  private final List<String> createdDatasetIds;
+  /* The dataset id used to create tables. */
+  private final String datasetId;
```
wchamber, line 18: Maybe add 'Must be non-null if {@link ResourceTracker#createTableCow} is used.`

PR #19, opened 2020-06-12
 fix the flaky test cased by latency metrics is not beeing tracked correctly in test

yonghaoy:
For the flaky test, I was lazy last time to assert increase because recordLatency is not called in too many places. The assert count equal just work all the time.
. But now, as we have more cows, and the metrics may get messed up with each other.


src/test/java/bio/terra/cloudres/util/MetricsHelperTest.java
```diff
@@ -76,22 +74,22 @@ public void testRecordErrorCount() throws Exception {
 
   @Test
   public void testRecordLatency() throws Exception {
+    // this is mapped to the Distribution defined in MetricsHelper, i.e.
+    // 0ms being within the first bucket & 1 ms in the 2nd.
```
wchamber, line 78: optional nit: The 0 & 1 bucket index here and on the assertions line 89-93 are supposed to have the same values. That gets a little confusing with 0ms and 1ms. maybe have a `final int zeroMsBucketIndex = 0` and `final int oneMsBucketIndex` and use those where appropriate? May be able to remove the /* parameter=*/ comments.

yonghaoy: done.

build.gradle
```diff
@@ -121,3 +121,11 @@ task minniekenny(type: Exec) {
 }
 
 test.dependsOn minniekenny
+
+// Suppress gradle javadoc errors and warnings
+tasks.withType(Javadoc) {
+    failOnError false
```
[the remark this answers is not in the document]
yonghaoy: before: see here:
`/home/runner/work/terra-cloud-resource-lib/terra-cloud-resource-lib/src/main/java/bio/terra/cloudres/common/TransformPage.java:8: error: unknown tag: F
/** Implementation for {@link Page} interface which supports transform Page<F> to Page<T> */`

After, all those errors are gone, able to build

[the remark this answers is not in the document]
wchamber: I know I was complaining before about it not letting me build for spotless code style, but ignoring javadoc errors is a little diferent and kind of extreeme. None of workspace manager, stairway, or data repo do that, so I think we should not ignore errors purely for consistency.

Can we do something to fix the Page<F> javadoc error? can you wrap it in {@code }  or just talk about F ?

build.gradle
```diff
@@ -126,4 +126,4 @@ artifactory {
             publishPom = true
         }
     }
-}
+}
```
wchamber: nit: mind adding the newline back?

PR #20, opened 2020-06-12
CA-825 Change directory structure to have gradle sub-project per CRL package

wchamber:
We want to publish and test CRL packages independently of each other to match the cloud api package division. This is the first step, reworking our directory structure to use top-level directories per CRL package and using Gradle multi-project builds.

Also deletes 'resourcemanager' package now that we have other examples and we don't plan on using that cloud library. We'll re-add it later with the api client library dependency.


wchamber:
> LGTM, just curious, what is the common practice to name them of google_bigquery, google_storage, instead of googelbigquery, googlestorage?

Dunno, that just seem to be what google was doing, e.g. https://github.com/googleapis/java-datastore
Looking at that again, I think I'll rename the directories to use hyphens instead of underscores.

yonghaoy:
That SG, but for packing name, I know that is not good to use dash:
https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html
https://stackoverflow.com/questions/27131302/java-package-with/27131331#:~:text=Some%20Package%20Naming%20Conventions%20from%20java%20Documentation&text=You%20will%20have%20to%20change,to%20use%20an%20underscore%20instead.

But I guess it is okay if that is just directory name since that is how Google does?

wchamber:
Yep, just the directory, not the package itself.

build.gradle
```diff
@@ -30,55 +30,74 @@ gradle.taskGraph.whenReady { taskGraph ->
...
-    compile 'com.google.auth:google-auth-library-oauth2-http:0.20.0'
+    apply plugin: 'java-library'
+    apply plugin: 'java-test-fixtures'
+    apply plugin: 'com.diffplug.gradle.spotless'
 
-    compile 'org.slf4j:slf4j-api:1.7.28'
-    compile 'org.slf4j:slf4j-nop:1.7.25'
+    dependencies {
+        api 'com.google.auth:google-auth-library-oauth2-http:0.20.0'
+        api 'com.google.cloud:google-cloud-core-http:1.93.5'
```
andy7i, line 45: From reviewing this, I've learned a bit more about Gradle and using `implementation` or `api` instead of `compile` and the difference between the first two. [This SO question](https://stackoverflow.com/questions/44493378/whats-the-difference-between-implementation-and-compile-in-gradle) has someone explain it as: 

>if your app depends on library x which itself depends on y,z. if you use implementation only x api will be exposed, but if you use api y,z also will be exposed.

Do we need the dependencies of the google libraries exposed? (Or am I misunderstanding something?)

wchamber: Good question. We do need to expose the google library dependency. For example, the StorageCow is constructed with the `com.google.cloud.storage.StorageOptions`. Our clients need to be using the same version of the google library.
More generally, there are many parts of the underlying cloud library that we're exposing directly, like the BucketInfos or Acls. that is a part of the API of these packages - CRL is not trying to abstract away or hide the underlying library entirely.

PR #21, opened 2020-06-15
CA-825 Add multi project publishing and a publish github workflow

wchamber:
Artifactory: https://broadinstitute.jfrog.io/broadinstitute/webapp/#/artifacts/browse/tree/General/libs-snapshot-local/bio/terra/cloud-resource-lib


build.gradle
```diff
@@ -51,9 +37,9 @@ subprojects {
 
         implementation("com.google.guava:guava:29.0-jre")
         testImplementation 'junit:junit:4.13'
-        testFixturesApi 'junit:junit:4.13'
+        testFixturesImplementation 'junit:junit:4.13'
```
yonghaoy, line 40: what is this? I didn't find testFixturesImplementation online. Do we still need testImplementation?

wchamber: The testFixtures plugin is for re-using test only fixture code. Similar to the java plugin, it comes in `api` and `implementation` flavors for whether the dependency of the testFixtures is public api, or an implementation detail. I think we don't need to expose it as an API, so implementation is fine.
https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures

.github/workflows/publish.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Pull Vault image
+        run: docker pull vault:1.1.0
+      # Currently, there's no way to add capabilities to Docker actions on Git, and Vault needs IPC_LOCK to run.
+      - name: Get Vault token
+        id: vault-token-step
+        run: |
+          VAULT_TOKEN=$(docker run --rm --cap-add IPC_LOCK \
+            -e "VAULT_ADDR=${VAULT_ADDR}" \
+            vault:1.1.0 \
+            vault write -field token \
```
marctalbott, line 26: Why is this a `vault write` instead of a `vault read`?

wchamber:
> The PR LGTM, just make sure
> https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/.github/workflows/build-and-test.yaml#L60
> is tracked somethere.

changed to TODO CA-814.

google-storage/gradle.properties
```diff
@@ -0,0 +1,2 @@
+group = bio.terra.cloud-resource-lib
+version = 0.0.0-SNAPSHOT
```
marctalbott, line 2: not entirely sure that I'm correctly following how these values are being used, but should there also be a `name` property in these files? I'm basing this on lines 88-90 in build.gradle

wchamber: it's default inferred from the directory for multi-project builds. So not necessary, but we could add it.

.github/workflows/publish.yml
wchamber: short answer: I'm cargo-culting how every other broad github action gets secrets out of vault. e.g. https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/.github/workflows/build-and-test.yaml#L26-L34
https://github.com/DataBiosphere/terra-workspace-manager/blob/master/.github/workflows/master_push.yml#L31-L42

longer answer: I think we're using github stored approle secrets to create a temporary vault token for our github approle account. This happens to be a `write` for no doubt good Vault reasons. https://www.vaultproject.io/docs/auth/approle. I don't know why we don't have a github action for doing this.

google-storage/gradle.properties
wchamber: These properties are being used in the magic gradle `subprojects` closure to get group/version for the Maven publication for each subproject. https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/ccc9bf5afd8adcfd8770554ce431f7d23c79f17e/build.gradle#L88 (see enclosing `subprojects { subproject ->` on L19.

.github/workflows/publish.yml
marctalbott: 👍 I figured it might just be common to how we get secrets out of vault, but the write stuck out to me among the other reads so wanted to check!

PR #22, opened 2020-06-16
Support BigQuery query and insert. also supports throwing checked exception from Cows

yonghaoy:
https://broadworkbench.atlassian.net/browse/CA-831


common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -39,6 +39,29 @@ public OperationAnnotator(ClientConfig clientConfig, Logger logger) {
...
+
+  /**
+   * Executes the CowOperation and allows for checked exceptions..
+   *
+   * @param cloudOperation: the {@link CloudOperation} to operate.
+   * @param cowExecute: how to execute this cloud operation
+   * @param cowSerialize: how to serialize request
+   * @return the result of executing the {@code cowOperation}
+   */
+  public <R, E extends Exception> R executeCowOperationCheckedException(
```
wchamber, line 60: nit: maybe `executeCheckedCowOperation`

common/src/main/java/bio/terra/cloudres/common/CloudOperation.java
```diff
@@ -23,13 +23,15 @@
   // Google BigQuery Operations
   GOOGLE_CREATE_DATASET,
   GOOGLE_CREATE_BIGQUERY_TABLE,
+  GOOGLE_INSERT_BIGQUERY,
```
wchamber, line 26: should this be `GOOGLE_INSERT_BIGQUERY_TABLE` to match `GOOGLE_QUERY_BIGQUERY_TABLE`.
as-is, I'm not sure what's being inserted where.

google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/SerializeUtils.java
```diff
@@ -70,4 +70,29 @@ static JsonObject convert(
...
+    JsonObject jsonObject = new JsonObject();
+    jsonObject.add("queryJobConfiguration", gson.toJsonTree(queryJobConfiguration));
+    jsonObject.add("jobId", gson.toJsonTree(jobId));
+    jsonObject.add("jobOptions", gson.toJsonTree(jobOptions));
+    return jsonObject;
+  }
+
+  static JsonObject convert(InsertAllRequest insertAllRequest) {
+    JsonObject jsonObject = new JsonObject();
+    jsonObject.add("insertAllRequest", new Gson().toJsonTree(insertAllRequest));
```
wchamber, line 95: I think this could contain the information of what's being inserted in every row. I'm not sure we want to log that. maybe just the table id is sufficient. WDYT?
For QueryJobConfiguration, the query would also be fully logged. Do we want that?

google-bigquery/src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -150,4 +152,44 @@ public void listTables() {
...
+  public void insertThenQuery() throws Exception {
+    String fieldName = "field1";
+    String fieldValue = "value1";
+
+    String datasetId =
+        resourceTracker.createDatasetCow().getDatasetInfo().getDatasetId().getDataset();
+    TableId tableId = TableId.of(datasetId, "table1");
+    // The random UUID tableId is not the standard table id format, to make query work, manually
+    // creates shorter name.
+    TableCow tableCow = resourceTracker.createTableCow(tableId);
```
wchamber, line 166: what's going on here? why can't we just use `createTableCow()`?

yonghaoy: the UUID is too long for tableId and is not a good table id format. The query syntax can not recognize that complicated tableId:
`select field form tableId`.
And I get some syntax errors saying 
Syntax error: Missing whitespace between literal and alias at, something like that.

google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/SerializeUtils.java
yonghaoy: For insertRequest, changed to just log the table id and number of rows to insert.
for QueryJobConfiguration, will the query too large to log? I thought they will just be some simple query strings, and we should log them. 
Is that okay?

google-bigquery/src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
wchamber: That sounds like we should be escaping the tableId string. Maybe that's too much work for this test?
What if we change the new `createTableCow` to just take a `String tableName`? I think then we could have the no argument `createTableCow` method call the `createTableCow(String tableName)` method. Then we wouldn't need to create this ad-hoc dataset and mess around with getting tableids. All we're doing is saying we want to specify the table name instead of random, that let's us do it.

google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/SerializeUtils.java
wchamber: that sounds good to me.

google-bigquery/src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
yonghaoy: Just passing a tableName is my initial version, the reason I choose this way is that datasetId also need to be escaping in some cases. 
My previous version of query code is 
`select field form datasetid.tableId`, and datasetId is also not a good format. It works if I create a simple dataset id as well. 
It works now because I remove the `datasetId` from query and put it under `.setDefaultDataset(datasetId)`. Not sure if we can do it all the time so I just let tracker take entire tableId. 
I think what we can do now is change this method to`createTableCow(String tableName)`, and if some tests need a simple dataetId name, they will need to create their own tracker and do tearDown themself. 
Or if you think we won't need the datasetId appears in query string, I can do ahead and change `createTableCow(String tableName)`.

Thanks

wchamber: That sounds more and more like we should just be escaping things properly in the query code and not worrying about what form our dataset and table ids are.
Take a look at https://cloud.google.com/bigquery/docs/parameterized-queries?

common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -133,9 +156,21 @@ private JsonObject createExceptionEntry(Exception executionException) {
     R execute();
   }
 
+  /** How to execute this operation. Like {@link CowExecute}, but allows for checked exceptions. */
+  @FunctionalInterface
+  public interface CowExecuteCheckedException<R, E extends Exception> {
```
wchamber, line 161: maybe `CowCheckedExecute`? little shorter and matches `executeCheckedCowOperation`

common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -39,6 +39,29 @@ public OperationAnnotator(ClientConfig clientConfig, Logger logger) {
    */
   public <R> R executeCowOperation(
       CloudOperation cloudOperation, CowExecute<R> cowExecute, CowSerialize cowSerialize) {
+    try {
+      return executeCheckedCowOperation(
+          cloudOperation,
+          (CowExecuteCheckedException<R, BogusException>) cowExecute::execute,
```
wchamber, line 45: nit: maybe worth a tiny comment here
`// Wrap cowExecute in a CowExecuteCheckedException so we can use the same code even though it will never throw a checked exception.`

google-bigquery/src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
wchamber: Ok, sorry, that sounds good if we can't escape. Sorry for the pad suggestion on parameterized queries which doesn't work for dataset/table ids.
Last attempt at escaping ```"SELECT " + fieldName + " FROM `" + datasetId + "`.`" + tableName +"`"```

google-bigquery/src/test/java/bio/terra/cloudres/google/bigquery/SerializeUtilsTest.java
```diff
@@ -98,4 +98,37 @@ public void convertTableIdWithTableDefinitionWithTableOption() {
                 BigQuery.TableOption.fields())
             .toString());
   }
+
+  @Test
+  public void convertQueryConfigWithJobOption() {
+    assertEquals(
+        "{\"queryJobConfiguration\":{\"query\":\"query\",\"positionalParameters\":[],\"namedParameters\":{},"
+            + "\"defaultDataset\":{\"dataset\":\"datasetId1\"},\"useLegacySql\":false,\"type\":\"QUERY\"},"
+            + "\"jobOptions\":[{\"rpcOption\":\"FIELDS\",\"value\":\"jobReference,configuration\"}]}",
```
andy7i, line 107: this makes me wish for scala's `"""strings with "quotes" like this"""` :(

google-bigquery/src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -150,4 +152,44 @@ public void listTables() {
...
         bigQueryCow.listTables(reusableDataset.getDatasetId().getDataset()));
   }
+
+  @Test
+  public void insertThenQuery() throws Exception {
+    String fieldName = "field1";
+    String fieldValue = "value1";
+
+    String datasetId =
+        resourceTracker.createDatasetCow().getDatasetInfo().getDatasetId().getDataset();
```
andy7i: These look like integration tests? Do we have a plan for datasets/tables that aren't cleaned up?

google-bigquery/src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
yonghaoy: done

google-bigquery/src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
yonghaoy: eventually all those resources will be cleaned up by the janitor service we are building. But for now, we manually tacking all resources using BigQueryResourceTracker, and clean up inside it.

common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -39,6 +39,29 @@ public OperationAnnotator(ClientConfig clientConfig, Logger logger) {
...
+   * Executes the CowOperation and allows for checked exceptions..
+   *
+   * @param cloudOperation: the {@link CloudOperation} to operate.
+   * @param cowExecute: how to execute this cloud operation
+   * @param cowSerialize: how to serialize request
+   * @return the result of executing the {@code cowOperation}
+   */
+  public <R, E extends Exception> R executeCheckedCowOperation(
+      CloudOperation cloudOperation,
+      CowExecuteCheckedException<R, E> cowExecute,
```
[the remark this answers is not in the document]
yonghaoy: done

common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -39,6 +39,29 @@ public OperationAnnotator(ClientConfig clientConfig, Logger logger) {
    */
   public <R> R executeCowOperation(
       CloudOperation cloudOperation, CowExecute<R> cowExecute, CowSerialize cowSerialize) {
+    try {
+      return executeCheckedCowOperation(
+          cloudOperation,
+          (CowExecuteCheckedException<R, BogusException>) cowExecute::execute,
+          cowSerialize);
+    } catch (BogusException e) {
+      throw new AssertionError("Our BogusException should never be thrown by cowExecute.", e);
```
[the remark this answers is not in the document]
yonghaoy: currently executeCowOperation call executeCheckedCowOperation and use the BogusException, because for checked one it expects to have the try-catch block for checkedException. So the other way around won't work because it can not do that

google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
```diff
@@ -123,4 +123,31 @@ public TableCow getTable(String datasetId, String tableId, TableOption... tableO
...
     return listTables(DatasetId.of(datasetId), tableListOptions);
   }
+
+  /** See {@link BigQuery#query(QueryJobConfiguration, BigQuery.JobOption...)}. */
+  public TableResult query(QueryJobConfiguration configuration, BigQuery.JobOption... jobOptions)
+      throws InterruptedException {
+    return operationAnnotator.executeCheckedCowOperation(
+        CloudOperation.GOOGLE_QUERY_BIGQUERY_TABLE,
+        () -> bigQuery.query(configuration, jobOptions),
+        () -> convert(configuration, jobOptions));
```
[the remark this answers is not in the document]
wchamber: CowExecuteCheckedException is a "functional interface", it's an interface that's equivalent to a single function. Java 8+ allows lambdas `() ->` to automagically be inferred to implement the interface, if the type's line up. Here, that lambda matches the functional interface of CowExecuteCheckedException.
More explanation around googling for " java functional interface lambda expression"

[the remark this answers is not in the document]
yonghaoy: We can cast it here to add `(OperationAnnotator.CowCheckedExecute<TableResult, InterruptedException>)`
But this is redudent because bigQuery.query() method alreads decliare it returns TableResult and trhows  InterruptedException, so the compliers how to map it. But for the executeCowOperation, we need to cast it there to let it know the method `cowExecute` throws a BogusException, then the line knows it, we can have the try catch block.

PR #25, opened 2020-06-19
CA-867 Add openapi initial definition for cloud resources UID and stubs for recording created resources for cleanup 

wchamber:
Implement StorageCow create bucket recording as an example with test.


common/src/main/java/bio/terra/cloudres/common/cleanup/NullCleanupRecorder.java
```diff
@@ -0,0 +1,11 @@
+package bio.terra.cloudres.common.cleanup;
+
+import bio.terra.cloudres.resources.CloudResourceUid;
+
+/** Null object for {@link CleanupRecorder} to do nothing. */
+public class NullCleanupRecorder implements CleanupRecorder {
```
yonghaoy, line 6: Noop or NOP or Null?
I know this:
https://github.com/qos-ch/slf4j/blob/master/slf4j-nop/src/main/java/org/slf4j/nop/NOPLogger.java

README.md
```diff
@@ -61,3 +71,5 @@ use. To add a new CRL package to support a new client library:
 4. Add the gradle project for the new package to the top [`settings.gradle`](settings.gradle).
 5. Add the gradle project for the new package to the list of `artifactory` `publications` in the top
 [`build.gradle`](build.gradle)
+
+TODO add instructions/considerations for adding a new Cloud Object Wrapper or cloud resource.
```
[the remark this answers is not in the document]
wchamber: I kind of want one checklist of everything that's involved in setting up new stuff. That'll touch the openapi, adding cleanup mode calls, operation annotators, Cows, etc. So even though it will have one thing for cleanup mode, worth keeping separate I think,.

common/src/main/java/bio/terra/cloudres/common/ClientConfig.java
```diff
@@ -2,27 +2,59 @@
...
+   * The {@link CleanupRecorder} to use to record resources for clenaup, if running in cleanup mode.
+   */
+  public CleanupRecorder getCleanupRecorder() {
+    return cleanupRecorder;
+  }
+
+  /** Record a cloud resource to be created for cleanup, if running in cleanup mode. */
+  public void recordForCleanup(CloudResourceUid resource) {
+    if (!getCleanupConfig().isPresent()) {
+      return;
```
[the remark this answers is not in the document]
wchamber: I think it's ok ot silently return. In production code, we'll hit this return, and that's expected.

common/src/main/java/bio/terra/cloudres/common/cleanup/CleanupConfig.java
```diff
@@ -0,0 +1,34 @@
...
+ * Configuration for CRL running in cleanup mode, where it tracks cloud resources being created and
+ * cleans them up later.
+ *
+ * <p>See the README on Cleanup Mode.
+ */
+@AutoValue
+public abstract class CleanupConfig {
+
+  /** An id to differentiate this run of CRL in cleanup mode from other runs. Any format allowed. */
+  public abstract String cleanupId();
```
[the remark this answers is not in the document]
wchamber: Adding more to the comment to explain what it's for. It is passed by the client, as CleanupConfig is a part of the ClientConfig.
There's TODO in IntegrationUtils about setting different ids for CRL integration tests.

common/src/main/java/bio/terra/cloudres/common/cleanup/NullCleanupRecorder.java
wchamber: Don't have a strong preference on the name. Always heard of it as the "null object pattern" https://en.wikipedia.org/wiki/Null_object_pattern

common/src/main/java/bio/terra/cloudres/common/ClientConfig.java
```diff
@@ -2,27 +2,59 @@
...
+
+/** Configuration class to manage CRL behavior. */
 public class ClientConfig {
   private final String clientName;
+  private final Optional<CleanupConfig> cleanupConfig;
+  private final CleanupRecorder cleanupRecorder;
 
-  private ClientConfig(String clientName) {
+  private ClientConfig(
+      String clientName, Optional<CleanupConfig> cleanupConfig, CleanupRecorder recorder) {
```
yonghaoy, line 18: So the when clients create a Cow, they will need to create a cleanUpConfig as well as a recorder? I feel that is too much/confusing/unnessary for client, can we remove the setRecorder from builder and we create recorder for them? We want to have a central point for client to put all cleanup related stuff.  All cleanUp related things should be passed in cleanupConfig.  
e.g.:

```
ClientConfig.newBuilder().setClientUpConfig(CleanupConfig.builder()
          .setTimeToLive(Duration.ofHours(2))
          .setCleanupId("id")
          .setRecorderType(IN_MEMORY)
          .build()).build();
```

Then in our ClientConfig.build() method,
we can have:  `recorder = createCleanupRecorder(recorderType/(or cleanupConfig));`

If we can do that way, that is less confusing and the clientConfig is easier for client to use.

wchamber: When they create the ClientConfig, they need to create the CleanupConfig and the CleanupRecorder if they want to have things cleaned up. I imagine most clients will re-use the same ClientConfig for multiple Cows, not create it everytime. Or even if they do, they'll have a function for creating the CleanupConfig.

We don't need to create a CleanupConfig for the in memory recorder by default. We don't want any recording to happen by default - production doesn't do the recording.

I'm not sure what the best way to create the CleanupRecorder is. You're suggesting create it with a static factory function `createCleanupRecorder` in ClientConfig.build(). There's also the matter of the client needing to set up the right CleanupRecorder when it's going to be initialized with properties like which pub/sub queue/http endpoint to hit. Creating it with a static factory function also makes testing more difficult because it's hard to acccess the CleanupRecorder being used. While there's more boilerplate in test setup, using dependency injection to insert the CleanupRecorder to the ClientConfig is more general.
Maybe talk offline?

wchamber: Thanks. Instead of using an enum to figure out which CleanupRecorder to use, made a locater service to grab the singular right one. I'm imagining for the real service we have some configuration saying what endpoint to hit for the recorder. Once we have that, also setting the enum in the ClientConfig feels redundant. LMK what you think.

common/src/main/java/bio/terra/cloudres/common/ClientConfig.java
```diff
@@ -2,27 +2,44 @@
...
+  /**
+   * The {@link CleanupConfig} of how created resources should be cleaned up, or empty if not
+   * running in cleanup mode.
+   */
+  public Optional<CleanupConfig> getCleanupConfig() {
+    return cleanupConfig;
+  }
+
+  /** Record a cloud resource to be created for cleanup, if running in cleanup mode. */
+  public void recordForCleanup(CloudResourceUid resource) {
```
yonghaoy, line 36: it looks strange that we have a public method for recordForCleanup. That means clients is able to manually call this to record things? I feel this should be logic happens automatically/inside in Cows's creating resource method?  CRL should take care of this if in test mode. 
My understanding of the logic would be: 
ClientConfig:

```
ClientConfig.newBuilder()
.setClientName(name)
.setCleanupConfig(
CleanupConfig.newBuilder()
  .setIsTestMode(true)
  .setJanitorServiceAddress().
  .setRecoderType(IN_MEMORY).build()
```
Then in cow's create method, we can have 

```
/** See {@link Storage#create(BlobInfo, Storage.BlobTargetOption...)}. */
  public BlobCow create(BlobInfo blobInfo) {
    Blob blob =
        operationAnnotator.executeCowOperation(
            CloudOperation.GOOGLE_CREATE_BLOB,
            () -> storage.create(blobInfo),
            () -> SerializeUtils.convert(blobInfo)),
            /*isTest */true,
    return new BlobCow(clientConfig, blob);
  }
```
OperationAnnotator holds the cleanupRecoder: 

```
  public OperationAnnotator(ClientConfig clientConfig, Logger logger) {
    this.clientConfig = clientConfig;
    this.logger = logger;
    this.recoder = new CleanupRecoder(clientConfig.getCleanupConfig);
  }

public <R> R executeCowOperation(..... isTest)
  if(isTest) {
  recoder.record();
}
```
But that seems it's hard to generate the reourceUid in OperationAnnotator? An alternative way is to move recorder and record to cows. 
So in cows's create method, we can have 

```
public BucketCow create(BucketInfo bucketInfo) {
   recoder.record(new GoogleBucketUid().bucketName(bucketInfo.getName()))
    Bucket bucket =
        operationAnnotator.executeCowOperation(
            CloudOperation.GOOGLE_CREATE_BUCKET,
            () -> storage.create(bucketInfo),
            () -> SerializeUtils.convert(bucketInfo));
    return new BucketCow(clientConfig, bucket);
  }
```
In that way, we will get fully control about the record parts and all magic will happens inside CRL, no change in client's test code. 
What do you think? Happy to chat more!

wchamber: I don't like putting the recording in the OperationAnnotator bbecuase it's only related to some of the operations. We do start needing to always generate the resourceUid, and I don't think that can be done automatically from the OperationAnnotator.

Not sure what you're suggesting for moving the recorder to be in the Cow. The recording is happening within the Cow's create method. Are you suggesting we inject the recorder into the Cow instead of getting it from the Locator?
Something has to be different in client test code to figure out if recording should happen. There also has to be something in the client test configurtion to tell the recorder where to record things. Maybe let's talk offline.

wchamber: Per offline talk, tried again. This is more like creating a new CleanupRecorder with every cow operation (your second to last code block) but hides whether a new instance is being created behind a static method.
With this htinking, we'd add things like the janitor http address to the CleanupConfig and make that the only cleanup configuration, as you're suggesting. There's only one implementation of the CleanupRecorder, at least so far as any Cow knows.

PR #27, opened 2020-07-06
CA-868 Add cleanup recording for BigQuery datasets and tables.

wchamber:


common/src/main/resources/cloud_resources_uid.yaml
```diff
@@ -10,6 +10,37 @@ paths: {}
...
+        - tableName
+      properties:
+        resourceType:
+          type: string
+          default: googleBigQueryTableUid
+        projectId:
+          type: string
+        datasetId:
+          type: string
+        tableName:
```
yonghaoy, line 42: should this be tableId as well?

wchamber: done

PR #28, opened 2020-07-06
[CA 873] create separate package for cloud-resource-schema

yonghaoy:
Seems that is the best and easiest way of doing that:

- Cloud Resource Schema can be used in multiple place which also be part of CRL, so it should be in a CRL seperate package.
- Split if from /common because we want to avoid cycle dependeny with the usage of janitor service.
- Also, this is the easiest way if we want to completly move cloud-resource-schema out of CRL in the future.

Package:
https://broadinstitute.jfrog.io/broadinstitute/webapp/#/artifacts/browse/tree/General/libs-snapshot-local/bio/terra/cloud-resource-lib/cloud-resource-schema


yonghaoy:
Ready for another review. Thanks

README.md
```diff
@@ -69,7 +69,33 @@ use. To add a new CRL package to support a new client library:
...
+Gradle:
+```
+implementation(group: 'bio.terra.cloud-resource-lib', name: 'google-bigquery', version: '0.0.0-SNAPSHOT')
+```
+### cloud-resource-schema
+The general schema for how cloud resources are presented in CRL. 
+
+Gradle:
+```
+implementation(group: 'bio.terra.cloud-resource-lib', name: 'cloud-resource-schema', version: '0.0.0-SNAPSHOT')
```
wchamber, line 100: maybe explain the pattern and where to find what packages there are instead of adding bullets for each package. Or maybe it's better to have space for each of them. WDYT?

.github/workflows/publish.yml
```diff
@@ -47,6 +47,8 @@ jobs:
           echo ::add-mask::$ARTIFACTORY_PASSWORD
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
+      - name: Generate Cloud Resource Schema using openapi generator
+        run: ./gradlew openApiGenerator java
```
wchamber, line 51: Why does it matter if we generate the schema in the github action for publishing the artifact? Anything needed by the jars should be a dependency.

README.md
yonghaoy: changed the style a little bit. PTAL, thanks!

PR #29, opened 2020-07-16
[CA 873] Modify the schema not to use interface pattern.

yonghaoy:


cloud-resource-schema/src/main/resources/cloud_resources_uid.yaml
```diff
@@ -44,54 +36,40 @@ components:
...
       required:
-        - resourceType
         - projectId
       properties:
-        resourceType:
-          type: string
-          default: googleProjectUid
         projectId:
           type: string
+    # We are not doing polymorphism at this moment because of lack of support in swagger-codegen or openApiGenerator
```
wchamber, line 60: Maybe a little before this (which should have been there earlier, sorry).
`# A cloud resource unique identifier. Each CloudResourceUid represents exactly one cloud resource.`

cloud-resource-schema/src/test/java/bio/terra/clouders/resources/CloudResourceUidTest.java
```diff
@@ -31,36 +31,11 @@ public void assertSerializationIdempotency(CloudResourceUid resource, Class<?> r
...
-  public void googleBucket() throws Exception {
-    GoogleBucketUid bucket = new GoogleBucketUid().bucketName("my-bucket");
-    assertSerializationIdempotency(bucket, GoogleBucketUid.class);
-  }
-
-  @Test
-  public void googleProject() throws Exception {
-    GoogleProjectUid project = new GoogleProjectUid().projectId("my-project");
-    assertSerializationIdempotency(project, GoogleProjectUid.class);
+    assertSerializationIdempotency(dataset, CloudResourceUid.class);
```
wchamber, line 39: I don't think we need to assert the class instance of since we're always using CloudResourceUid classes.
I still think there's some worth in making a test per resource type to make sure we can do what we want. That gives us more confidence when updating versions of the generator.

yonghaoy: Added those test back and use Class.getMethod to verify it. Is that okay?

cloud-resource-schema/src/test/java/bio/terra/clouders/resources/CloudResourceUidTest.java
```diff
@@ -18,49 +16,57 @@
...
+  public void assertSerializationIdempotency(CloudResourceUid resource, String getMethodName)
+      throws Exception {
     String serialized = objectMapper.writeValueAsString(resource);
     CloudResourceUid deserialized = objectMapper.readValue(serialized, resource.getClass());
     // Asserts that serializing->deserializing a {@link CloudResourceUid} yields an "equals"
     // instance.
     assertEquals(resource, deserialized);
-    // Assert that the deserialized instance is an instance of the same class.
-    assertThat(resource, Matchers.instanceOf(resourceClass));
+    assertNotNull(CloudResourceUid.class.getMethod(getMethodName));
```
wchamber, line 26: I think this just asserts that there is a method on CloudResourceUid with the method name. Since we're already setting the value in creating the CloudResourceUid, I'm not sure why we need that. I'm not worried that the code generation hs a way to retrieve values. assertEquals(resource, deserialized) is enough for me.

PR #30, opened 2020-07-21
[PF 22] Integrate with Janitor when create resources in test. 

yonghaoy:
https://broadworkbench.atlassian.net/browse/PF-22

1. JanitorService will be created per cleanup recorder
2. Remove all TestRecord stuff
3. A mock http JanitorService using WireMock
4. Switch to Janitor's CloudResourceUid instead of CRL-Schema

For now, add a Mock http client, and I am expecting we can call real janitor service once we solve the authZ problem


common/src/main/java/bio/terra/cloudres/common/cleanup/JanitorService.java
```diff
@@ -0,0 +1,25 @@
+package bio.terra.cloudres.common.cleanup;
+
+import bio.terra.janitor.ApiClient;
+import bio.terra.janitor.ApiException;
+import bio.terra.janitor.controller.JanitorApi;
+import bio.terra.janitor.model.CreateResourceRequestBody;
+
+/** An interface for Janitor service for cleanup resources in test. */
```
wchamber, line 8: nit: comment claims interface while it's a class

common/src/main/java/bio/terra/cloudres/common/cleanup/JanitorService.java
```diff
@@ -0,0 +1,25 @@
...
+    ApiClient client = new ApiClient();
+    client.setAccessToken(accessToken);
+    this.janitorApi = new JanitorApi(client.setBasePath(janitorBasePath));
+  }
+
+  public void createTrackedResource(CreateResourceRequestBody body) {
+    try {
+      janitorApi.createResource(body);
+    } catch (ApiException apiException) {
+      throw new RuntimeException("Failed to create tracked resource in Janitor", apiException);
```
wchamber, line 22: probably worth adding a specific exception type for this since if it happens during an integration test we want to be very clear it's our fault and not something in the service.

common/src/testFixtures/java/bio/terra/cloudres/testing/MockJanitorService.java
```diff
@@ -0,0 +1,68 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.github.tomakehurst.wiremock.WireMockServer;
+import com.github.tomakehurst.wiremock.core.WireMockConfiguration;
+import com.github.tomakehurst.wiremock.matching.RequestPattern;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class MockJanitorService {
```
wchamber, line 17: I don't really understand why we want to mock http connections. This feels like it's verifying the generated JanitorApi class instead of our use of it. 

I'm not sure what we're testing by going through the trouble of mocking http connections and importing this third-party wiremock library. What are we trying to test?
Is it the JanitorService is set up correctly and we can create the client with the right access tokens? I don't think there's much behavior to test there (maybe a unittest for creating the request body correctly), but if so, we only need to examine the http being sent for that unittest, not in every CowTest.
Is it that the CleanupRecorder is being called? That's what the recorded resources already present in the CowTest are asserting.

yonghaoy: The tests a correct request is sent with correct JanitorApi format.
(1) JanitorApi part we verified the access token is a valid Oauth token and there is no missing field. 
(2) We also verified the CloudResourceUid in request is correct, just like your TestRecord.

common/build.gradle
```diff
@@ -1,10 +1,11 @@
...
-    implementation project(':cloud-resource-schema')
-
     implementation "com.google.auto.value:auto-value-annotations:1.7.3"
     annotationProcessor "com.google.auto.value:auto-value:1.7.3"
     implementation 'io.opencensus:opencensus-api:0.26.0'
     implementation 'io.opencensus:opencensus-impl:0.26.0'
+    implementation "org.glassfish.jersey.inject:jersey-hk2:2.31"
 
     testImplementation 'com.google.cloud:google-cloud-resourcemanager:0.117.2-alpha'
+    testFixturesImplementation "com.github.tomakehurst:wiremock-jre8:2.27.0"
```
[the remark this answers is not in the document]
yonghaoy: I don't think other services have such cases. Spring should have some easier way to do this.
I know WM's test use @MockBean annotation and create a mock SAM instance. 
i am not sure if we should pull wiremock, did some search and compare, seems this is one of the best choice if we want to mock a http service?

common/src/testFixtures/java/bio/terra/cloudres/testing/MockJanitorService.java
wchamber: That can be used on a single unittest for the JanitorService and not for every CowTest. The CowTests just need to check that they're calling CleanupRecorder correctly. They don't need to test that CleanupRecorder calls the Janitor correctly.

PR #31, opened 2020-07-22
[PF 22] Migirate to use the Janitor generated CloudResourceUid

yonghaoy:
Ticket: https://broadworkbench.atlassian.net/browse/PF-22


cloud-resource-schema/README.md
```diff
@@ -0,0 +1,5 @@
+# Cloud Resource Schema
+The unified schema to represent cloud resources. It defines the schema in OpenApi format and also generates Java code 
+based on the schema. Both OpenApi schema and Java code in this package. You can either use the Java code directly or 
+generated your own code from the OpenApi schema.
+Currently CRL Janitor only uses this OpenApi yaml file and generates its own Java code.
```
wchamber, line 5:
```
# Cloud Resource Schema
This package defines a unified, serializable schema to represent all cloud resources interesting to Terra. These
cloud resource ids should uniquely identify a resource according to its cloud provider.

It defines the schema in OpenApi format and also generates Java code 
based on the schema. Both OpenApi schema and Java code are in this package. You can either use the Java code directly or 
generated your own code from the OpenApi schema.

Currently, the terra-cloud-resource-lib does not use this directly. It instead uses the client library generated by
the Terra Janitor that includes generated code using this schema.
```

common/build.gradle
```diff
@@ -1,10 +1,10 @@
 dependencies {
-    implementation project(':cloud-resource-schema')
-
     implementation "com.google.auto.value:auto-value-annotations:1.7.3"
     annotationProcessor "com.google.auto.value:auto-value:1.7.3"
     implementation 'io.opencensus:opencensus-api:0.26.0'
     implementation 'io.opencensus:opencensus-impl:0.26.0'
+    implementation "org.glassfish.jersey.inject:jersey-hk2:2.31"
```
wchamber, line 6: do we need this?

yonghaoy: Let me remove this. We will need this while we port JanitorApi.

PR #32, opened 2020-07-24
[PF 22]: Make a Janitor call when creating resource with unit test

yonghaoy:


build.gradle
```diff
@@ -48,6 +48,7 @@ subprojects { subproject ->
         testFixturesImplementation 'org.junit.jupiter:junit-jupiter-api:5.6.2'
         testImplementation 'org.hamcrest:hamcrest:2.2'
         testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.6.2'
+        testImplementation group: 'org.glassfish.jersey.core', name: 'jersey-client', version: "2.31"
```
wchamber, line 51: do we need these? Why isn't what we need included from janitor client library? (ditto in common/build.gradle).

common/src/main/java/bio/terra/cloudres/common/JanitorException.java
```diff
@@ -0,0 +1,8 @@
+package bio.terra.cloudres.common;
+
+/** Exception thrown when Janitor call throws exception. */
```
wchamber, line 3:
```
Exception thrown when Janitor call to record resource for cleanup fails.
<p> There is nothing for the CRL client to do about this. This should hopefully happen very rarely, or we should revisit.
```

common/src/main/java/bio/terra/cloudres/common/cleanup/CleanupConfig.java
```diff
@@ -23,6 +23,15 @@
   /** How long created resources should live (at least) before being cleaned up. */
   public abstract Duration timeToLive();
 
+  /**
+   * The client access token when calling Janitor, e.g. when using Google SA account credentials, it
+   * would be {@code serviceAccountCredentials.refreshAccessToken().getTokenValue()}
```
wchamber, line 28: How long is the accessToken valid? Right now, this is passed in once for the lifetime of the ClientConfig, which could be a long time for the life of the server. Do we need to pass in the credentials and be able to refresh them instead? Is there a way to tell how long the token lives for and refresh as needed? Is there a google library to handle this for us?
https://googleapis.dev/java/google-auth-library/latest/com/google/auth/oauth2/AccessToken.html#getExpirationTime--

common/src/main/java/bio/terra/cloudres/common/cleanup/CleanupRecorder.java
```diff
@@ -29,6 +43,26 @@ public static void record(CloudResourceUid resource, Optional<CleanupConfig> cle
     return testRecord.startNewRecord();
   }
 
+  /** Providers an {@link ApiClient}. */
+  public static void providerApiClient(ApiClient apiClient) {
```
wchamber, line 47: nit: 'provides' not 'providers' - comment and method name.
Does this need to be `public` or just @VisibleForTesting?

common/src/test/java/bio/terra/cloudres/common/cleanup/CleanupRecorderTest.java
```diff
@@ -26,34 +44,69 @@
...
     List<CloudResourceUid> record = CleanupRecorder.startNewRecordForTesting();
-    CleanupRecorder.record(RESOURCE_2, Optional.of(CLEANUP_CONFIG));
-    CleanupRecorder.record(RESOURCE_3, Optional.of(CLEANUP_CONFIG));
+    CleanupRecorder.record(RESOURCE_2, CLIENT_CONFIG);
+    CleanupRecorder.record(RESOURCE_3, CLIENT_CONFIG);
 
     assertThat(record, Matchers.contains(RESOURCE_2, RESOURCE_3));
+
+    // All record are captured.
+    assertApiClientSet(3);
```
wchamber, line 70: let's split these tests/asserts out from the newRecordForTesting tests.

common/src/test/java/bio/terra/cloudres/common/cleanup/CleanupRecorderTest.java
```diff
@@ -26,34 +44,69 @@
...
+    CleanupRecorder.record(RESOURCE_2, CLIENT_CONFIG);
 
     assertThat(record1, Matchers.contains(RESOURCE_1));
     assertThat(record2, Matchers.contains(RESOURCE_2));
+    assertApiClientSet(2);
+  }
+
+  @Test
+  public void createJanitorResource() {
+    // Not worth testing but fine to have.
```
wchamber, line 98: uh, if we don't think it's worth it, let's not have it. or just remove the comment

build.gradle
yonghaoy: The test is mocking the API client, and in spy method, it need to know the GenericType(a parameter in `apiClient.invoke()`) method. https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/GenericType.html
Api client also have the dependency.  But it is not complied unless I add this here.

common/src/main/java/bio/terra/cloudres/common/cleanup/CleanupConfig.java
yonghaoy: Access token expires 1 hr. My imagine the workflow in a test would be: 
```
public class test {
   BigQueryCow cow = new BigQueryCow(ClientConfig.setCleanupCofig(CleanpConfig.setAccessToken(ServiceAccountCreds.get/refrehAccessToken(Scope))));

  @Test  
  public void test() {
     cow.createTable(...)
   }
}
```

Now, the work is given to the client refresh the token if needed, if they want a new access token, they will need to create a new clientConfig. Now I think it is okay because in the process described above, 1 hr seems enough even they use a static Cow for the entire tests.
We can pass in the credentials, but the input would be `ServiceAccountCredentials` which is a Google credential only thing. What would you prefre?

common/src/main/java/bio/terra/cloudres/common/cleanup/CleanupRecorder.java
yonghaoy: Done rename. I think this need to be public and @VisibleForTesting, but it wont need to be public once we hooked up real janitor in Cow Tests

common/src/test/java/bio/terra/cloudres/common/cleanup/CleanupRecorderTest.java
yonghaoy: done

common/src/test/java/bio/terra/cloudres/common/cleanup/CleanupRecorderTest.java
yonghaoy: I removed the comments. Something not that worth to test but good to have..

common/src/main/java/bio/terra/cloudres/common/cleanup/CleanupConfig.java
wchamber: That's probably what a CRL integration test looks like, but it's not what running an ephemeral service looks like.
Imagine Workspace Manager being run in integration test mode. It probably has some spring configuration wiring to set up its ClientConfig. Then, it autowires the ClientConfig to whereever it needs a CRL.
Now this ephemeral environment might run for a few hours for a load test, and now our initial access token as expired.
I think we need to pass in the credentials. It is google specific at the moment, but our authn/authz to the janitor is google specific at the moment. Note that this doesn't mean that CRL needs to run in a Google environment, just that it gets the credentials.

yonghaoy: ah, oaky! Changed to credentials. PTAL, thanks!

common/src/test/java/bio/terra/cloudres/common/cleanup/CleanupRecorderTest.java
```diff
@@ -26,34 +47,73 @@
...
+            .resourceUid(RESOURCE_1)
+            .timeToLiveInMinutes(TTL_MIN)
+            .putLabelsItem("client", CLIENT_NAME)
+            .putLabelsItem("cleanupId", CLEANUP_ID),
+        CleanupRecorder.createJanitorResource(RESOURCE_1, CLIENT_CONFIG));
+  }
+
+  private void assertApiClientSet(int times) {
+    verify(spyApiClient, times(times)).setBasePath(JANITOR_PATH);
+    verify(spyApiClient, times(times)).setAccessToken(anyString());
```
yonghaoy, line 117: I changed to anyString(),because I feel no need to bother a token call to verify the token correct. 
But I also add something like credential.setscope.refreshToken,getValue, but I feel that is the same code block in main code... ptal

common/src/main/java/bio/terra/cloudres/common/cleanup/CleanupConfig.java
```diff
@@ -23,6 +25,18 @@
...
   public abstract Duration timeToLive();
 
+  /**
+   * The Google's {@link OAuth2Credentials} for accessing Janitor service. CRL refresh/generate
+   * access token using this credential.
+   *
+   * <p>Current Janitor only supports Google Credentials, and the most common usage would be a
+   * {@link com.google.auth.oauth2.ServiceAccountCredentials}
+   */
+  public abstract GoogleCredentials credentials();
```
wchamber, line 35: nit: feels a little funny to mention the super class OAuth2Credentials to then only allow the GoogleCredentials subclass. Maybe just pick one? The fact that we plan on setting up the janitor authz in a certain way doesn't feel like knowledge we should encode here - it belongs more in the deployment configuration of setting up the cleanupconfig.

common/src/main/java/bio/terra/cloudres/common/cleanup/CleanupRecorder.java
```diff
@@ -1,23 +1,52 @@
...
     }
+
+    CleanupConfig cleanupConfig = clientConfig.getCleanupConfig().get();
+    GoogleCredentials googleCredentials =
+        clientConfig.getCleanupConfig().get().credentials().createScoped(SCOPES);
+
+    try {
+      googleCredentials.refreshIfExpired();
+    } catch (IOException e) {
+      throw new RuntimeException(
```
wchamber, line 39: I think we should throw something more specific than a RuntimeException so that if this happens to a client it's even easier for them to figure out it's our cleanup code's fault. How about a JanitorAccessException to match our JanitorException, or re-use JanitorException?

common/src/test/java/bio/terra/cloudres/common/cleanup/CleanupRecorderTest.java
wchamber: lgtm

common/src/main/java/bio/terra/cloudres/common/cleanup/CleanupRecorder.java
wchamber: Still not `@VisibleForTesting`?

common/src/main/java/bio/terra/cloudres/common/cleanup/CleanupConfig.java
yonghaoy: Picked GoogleCredentials . ah, sorry, I tried to use OAuth2Credentials(that is still a google thing: google.auth. OAuth2Credentials). But it doesn't have refreshToken stuff. So I changed to GoogleCredentials.
For the knolwdge, Janitor repo will have some authZ explaination. While the knowledge encoded here is how to use Janitor via CRL. CRL, as a Janitor client, has its way to cordnicated with Janitor and offer a authZ solution about using Janitor in CRL.  And the comments tried to explain why CRL want it.

common/src/main/java/bio/terra/cloudres/common/cleanup/CleanupRecorder.java
yonghaoy: Use JanitorAccessException , and I changed JanitorException to JanitorApiException.

PR #36, opened 2020-08-13
[PF-29] Modify publish packages to let CRL client to use things under common/ package

yonghaoy:
Note that even we publish the common package, client don't need to put the dependency in their gradle. 
common would be auto imported if they have google-bigquery, google-storage in their dependency


wchamber:
Cool, thanks for figuring this out!

PR #39, opened 2020-08-21
PF-30 Add initial Cows for api-services resource manager Project create/delete

wchamber:


google-cloudresourcemanager/src/main/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCow.java
```diff
@@ -0,0 +1,82 @@
...
+import com.google.api.services.cloudresourcemanager.model.Operation;
+import com.google.api.services.cloudresourcemanager.model.Project;
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import java.io.IOException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** A Cloud Object Wrapper(COW) for Google API Client Library: {@link CloudResourceManager} */
+public class CloudResourceManagerCow {
```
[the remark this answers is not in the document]
wchamber: I think leaving the name as-is follows the existing Cow patterns: exactly suffix the class being wrapped with Cow.

google-cloudresourcemanager/src/main/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCow.java
```diff
@@ -0,0 +1,82 @@
...
+  private final Logger logger = LoggerFactory.getLogger(CloudResourceManagerCow.class);
+
+  private final OperationAnnotator operationAnnotator;
+  private final CloudResourceManager manager;
+
+  public CloudResourceManagerCow(
+      ClientConfig clientConfig, CloudResourceManager.Builder managerBuilder) {
+    operationAnnotator = new OperationAnnotator(clientConfig, logger);
```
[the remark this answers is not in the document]
wchamber: I think I'm inclined to keep accepting the `CloudResourceManager.Builder` in the constructor. This mirrors the `ServiceOptions.Builder` in the other Cows. While clients are likely to use default HttpTransports & JsonFactories, that's ok. So far, we haven't included the credentials in the ClientConfig.

google-cloudresourcemanager/src/main/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCow.java
```diff
@@ -0,0 +1,82 @@
...
+    manager = managerBuilder.build();
+  }
+
+  public Projects projects() {
+    return new Projects(manager.projects());
+  }
+
+  public class Projects {
```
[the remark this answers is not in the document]
wchamber: We talked offline about wrapping the API provided by the api-services libraries vs adding abstractions similar to the current cloud java libraries. leaving as wrapping for now to allow more client flexibility and less lift for adding new apis.

google-cloudresourcemanager/src/main/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCow.java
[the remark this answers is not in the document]
yonghaoy: I think it is okay not to include credentials  in client config, but we can have one more constructor like:
```
public CloudResourceManagerCow(
      ClientConfig clientConfig, GoogleCredentials credentials) {}
  ``` 
That's helps the clients which wants to just use the default settings, then we can have a helper method like:

```
private CloudResourceManager createCloudResourceManagerService(ClientConfig clientConfig, GoogleCredentials googleCredentials)
          throws IOException, GeneralSecurityException {
    HttpTransport httpTransport = GoogleNetHttpTransport.newTrustedTransport();
    JsonFactory jsonFactory = JacksonFactory.getDefaultInstance();

    return new CloudResourceManager.Builder(httpTransport, jsonFactory, new HttpCredentialsAdapter(googleCredentials))
            .setApplicationName(clientConfig.getClientName())
            .build();
  }
```

File: google-cloudresourcemanager/src/main/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCow.java [hunk 4]
```diff
@@ -0,0 +1,91 @@
...
+    private final CloudResourceManager.Projects projects;
+
+    private Projects(CloudResourceManager.Projects projects) {
+      this.projects = projects;
+    }
+
+    public Create create(Project project) throws IOException {
```
yonghaoy, line 46: add javadoc? ditto for `delete` class

google-cloudresourcemanager/src/test/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCowTest.java
```diff
@@ -0,0 +1,64 @@
...
+  public void createDeleteProject() throws Exception {
+    CloudResourceManagerCow managerCow = defaultManager();
+    String projectId = randomProjectId();
+    // TODO poll operation.
+    Operation operation =
+        managerCow
+            .projects()
+            .create(new Project().setProjectId(projectId).setParent(PARENT_RESOURCE))
+            .execute();
+    Thread.sleep(30000);
```
yonghaoy, line 56: do we need this tests sleep for 30 secs long?

google-cloudresourcemanager/src/test/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCowTest.java
```diff
@@ -0,0 +1,64 @@
...
+                    IntegrationCredentials.getAdminGoogleCredentialsOrDie()
+                        .createScoped(CloudResourceManagerScopes.all())))
+            .setApplicationName("crl-test"));
+  }
+
+  @Test
+  public void createDeleteProject() throws Exception {
+    CloudResourceManagerCow managerCow = defaultManager();
+    String projectId = randomProjectId();
+    // TODO poll operation.
```
yonghaoy, line 50: Is this todo here to indicate we will add 'get' inbetween to verify the created projects is correct?

google-api-services-common/src/main/java/bio/terra/cloudres/google/api/services/common/AbstractRequestCow.java
```diff
@@ -0,0 +1,55 @@
...
+      OperationAnnotator operationAnnotator,
+      AbstractGoogleClientRequest<T> request) {
+    this.operation = operation;
+    this.clientConfig = clientConfig;
+    this.request = request;
+    this.operationAnnotator = operationAnnotator;
+  }
+
+  /** See {@link AbstractGoogleClientRequest#execute()}. */
+  public T execute() throws IOException {
```
yonghaoy, line 36: Throws IOException in this abstract class seems little bit wired. 
Will all subclasses throw IOExceptions? How about different subs-class throws different exception?

google-cloudresourcemanager/src/main/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCow.java
```diff
@@ -0,0 +1,91 @@
...
+      public Create(CloudResourceManager.Projects.Create create, Project project) {
+        super(CloudOperation.GOOGLE_CREATE_PROJECT, clientConfig, operationAnnotator, create);
+        this.project = project;
+      }
+
+      @Override
+      protected Optional<CloudResourceUid> resourceUidCreation() {
+        return Optional.of(
+            new CloudResourceUid()
+                .googleProjectUid(new GoogleProjectUid().projectId(project.getProjectId())));
```
yonghaoy, line 62: One unrelated question, do we need to include parent id in cloudResourceUid?

google-cloudresourcemanager/src/test/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCowTest.java
```diff
@@ -0,0 +1,64 @@
...
+            .projects()
+            .create(new Project().setProjectId(projectId).setParent(PARENT_RESOURCE))
+            .execute();
+    Thread.sleep(30000);
+    managerCow.projects().delete(projectId).execute();
+  }
+
+  private static String randomProjectId() {
+    // Project ids must start with a letter and be no more than 30 characters long.
+    return "p" + IntegrationUtils.randomName().substring(0, 29);
```
yonghaoy, line 62: Do you think it is possible that this method will returns the same value as we substring 0 to 29?

google-api-services-common/README.md
```diff
@@ -0,0 +1,4 @@
+# Google Api Services Common
```
yonghaoy, line 1: Question:  what we want this package instead of adding more into the existing common/ library?  This depends on common/ and the package seems light-weighted. And packages might harder to manage if we can multiple common packages(the experience we have in Janitor when we forgot to upgrade the package version)

yonghaoy: And I feel the AbstractCow class is pretty generic, the only Google specific thing is `AbstractGoogleClientRequest`.
Is that possible to make a non-google specific common package to reduce the common package number and common abstract class number? for example, make `request` variable in that class more generic or write our own wrapper for a generic http request with `execute` method.
Even for Google, the network api are different, and we have to construct our own http request to make the call. Not sure what would be the longer plan for the package organization.

google-cloudresourcemanager/src/test/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCowTest.java
wchamber: this is hack to wait long enough for the create project operation to finish. I'm removing it in a follow up PR that polls until the create project operation is complete. See also next comment.

google-cloudresourcemanager/src/test/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCowTest.java
wchamber: The TODO is to poll the operation to figure out when the project has been created. We can add another get after that. (More PRs further down the road.)

google-api-services-common/src/main/java/bio/terra/cloudres/google/api/services/common/AbstractRequestCow.java
wchamber: So this is specifically to wrap `AbstractGoogleClientRequest#execute()` which always throws an IOException, so I think it's ok. Since subclasses of Google's `AbstractGoogleClientRequest` can't override their execute to throw other exceptions, we don't have to worry about that here.
Do you think a comment would be helpful?

google-cloudresourcemanager/src/main/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCow.java
wchamber: No. A project is uniquely identified by it's project id. A projects parent can change, but it will still be the same project.

google-cloudresourcemanager/src/test/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCowTest.java
wchamber: It's possible but not likely. It's returning a uuid, which are meant to cover the range of values evenly, not clump to similar values in the first 29 characters. There's a lot of randomness here.

google-api-services-common/README.md
wchamber: The AbstractCowRequest is meant to directly wrap a Google specific thing, the `AbstractGoogleClientRequest`. I'm not trying to create a general abstraction: this is still following the pattern of wrapping instead of abstracting. There's only the `execute` method today, but we could easily extend this for the other methods on `AbstractGoogleClientRequest`.

Since we will have multiple `com.google.api.services` packages, and they all use the same underlying patterns, like `AbstractGoogleClientRequest` I thought it made sense to have these as a separate package from the resource manager package. Our services API package will also use this. Right now, there's only the AbstractGoogleClientRequest, but we can define more common things, like a default json factory, or httptransport. In the next PR, operation utilities (again, targetting api.services operation patterns) are coming in.
I didn't want this to go into `common` because it is google specific. Whereas common should hopefully be re-used by the next clouds.

google-cloudresourcemanager/src/main/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCow.java
wchamber: getting a little verbose/boilerplate, but done.

google-cloudresourcemanager/src/main/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCow.java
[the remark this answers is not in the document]
wchamber: done, ptal

google-bigquery/gradle.properties
```diff
@@ -1,2 +1,2 @@
 group = bio.terra.cloud-resource-lib
-version = 0.0.9-SNAPSHOT
+version = 0.0.10-SNAPSHOT
```
yonghaoy, line 2: why we want to update this version?

yonghaoy (approved):

google-api-services-common/README.md
wchamber: AFAIK, all Google apis have api.services. I don't know which network ones you've been looking at. Here's something I found for cloud networks, not sure it's the right one. https://cloud.google.com/service-infrastructure/docs/apis

PR #41, opened 2020-08-27
[PF-25] Add support to get and delete dataset by DatasetId instead string name

yonghaoy:
Janitor needs this to delete/query dataset. But Janitor's Google Credential are in different project, so it needs the full DatasetId object which contains project id and dataset name. 

Also, change COW to return null if the cloud resource inside is null.

### DIFF
--- a/google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
+++ b/google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
@@ -91,22 +91,22 @@ public class BigQueryCow {
 
   /** See {@link BigQuery#getDataset(String, DatasetOption...)}. */
   public DatasetCow getDataSet(String datasetId, DatasetOption... datasetOptions) {
-    return new DatasetCow(
-        clientConfig,
+    Dataset rawDataset =
         operationAnnotator.executeCowOperation(
             CloudOperation.GOOGLE_GET_DATASET,
             () -> bigQuery.getDataset(datasetId, datasetOptions),
-            () -> convert(DatasetId.of(datasetId), datasetOptions)));
+            () -> convert(DatasetId.of(datasetId), datasetOptions));
+    return (rawDataset == null) ? null : new DatasetCow(clientConfig, rawDataset);
   }
 
   /** See {@link BigQuery#getDataset(DatasetId, DatasetOption...)}. */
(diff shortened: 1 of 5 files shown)


google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
```diff
@@ -81,14 +81,32 @@ public boolean delete(String datasetId, DatasetDeleteOption... deleteOptions) {
...
+  /** See {@link BigQuery#delete(DatasetId, DatasetDeleteOption...)}. */
+  public boolean delete(DatasetId datasetId, DatasetDeleteOption... deleteOptions) {
+    return operationAnnotator.executeCowOperation(
+        CloudOperation.GOOGLE_DELETE_DATASET,
+        () -> bigQuery.delete(datasetId, deleteOptions),
+        () -> convert(datasetId, deleteOptions));
+  }
+
   /** See {@link BigQuery#getDataset(String, DatasetOption...)}. */
   public DatasetCow getDataSet(String datasetId, DatasetOption... datasetOptions) {
```
gjuggler, line 93: opt / nit: I'm not sure how hard-set this function signature is, but the conditional null logic at the end suggests that maybe the return type here should be `Optional<DatasetCow>`. Or is DatasetCow itself supposed to gracefully handle a null-valued Dataset?

gjuggler, line 93: unrelated nit: this file should be consistent in whether it uses `dataSet` or `dataset`. The class name uses the latter capitalization, but the function uses the former. My instinct would be to align with Bigquery's capitalization in all usage.

(Given that this is a shared library, obviously in general we would want to be careful / intentional about making breaking name changes. But maybe an exception can be made if we're confident this isn't being actively used by any clients yet, and we're bumping the version anyway?)

yonghaoy: This follows the original cloud API:
https://googleapis.dev/java/google-cloud-bigquery/latest/com/google/cloud/bigquery/BigQuery.html#getDataset-com.google.cloud.bigquery.DatasetId-com.google.cloud.bigquery.BigQuery.DatasetOption...-
We feel it is easier for the client with cloud code knowledge to use CRL.

yonghaoy: ah, good catch! Thanks!

PR #42, opened 2020-09-01
PF-30 Add Operation utilities and method for Projects

wchamber:


google-api-services-common/src/main/java/bio/terra/cloudres/google/api/services/common/OperationUtils.java
```diff
@@ -0,0 +1,42 @@
...
+    Instant deadline = Instant.now().plus(timeout);
+
+    while (!isDone(operation.getOperationAdapter())
+        && operation.getOperationAdapter().getError() == null) {
+      if (Instant.now().plus(pollingInterval).isAfter(deadline)) {
+        throw new InterruptedException(
+            "Timeout during pollUntilComplete for operation "
+                + operation.getOperationAdapter().getName());
+      }
+      Thread.sleep(pollingInterval.toMillis());
```
yonghaoy, line 31: Is blocking thread a concern for us? do we need some unblocking approach like a scheduling service?

google-api-services-common/src/main/java/bio/terra/cloudres/google/api/services/common/OperationCow.java
```diff
@@ -0,0 +1,89 @@
...
+    return operationAdapter;
+  }
+
+  /**
+   * Execute a "get" for the Operation, returning a new OperationCow with the updated Operation
+   * information.
+   */
+  public OperationCow<T> executeGet() throws IOException {
+    AbstractRequestCow<T> getRequest = getRequestFactory.create(getOperation());
+    T newOperation = getRequest.execute();
```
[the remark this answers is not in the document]
wchamber: I don't think so. I think the `api.services` libraries always return the contents parsed as a class, that returned object won't be null.

google-api-services-common/src/main/java/bio/terra/cloudres/google/api/services/common/OperationUtils.java
wchamber: Data repo blocks their thread, and so far in MC Terra we've been working in a blocking paradigm. There are scalability cons to that approach, but until we switch to using `Future` or other async libraries, our clients are doing blocking so we're doing blocking. We can add more utility functions as needed.

yonghaoy (approved):

PR #44, opened 2020-09-02
PF-30 Add resource manager get project method

wchamber:


common/src/main/java/bio/terra/cloudres/common/CloudOperation.java
```diff
@@ -5,6 +5,9 @@
   // Google Resource Manager Operations
   GOOGLE_CREATE_PROJECT,
   GOOGLE_DELETE_PROJECT,
+  GOOGLE_GET_PROJECT,
+  GOOGLE_GET_IAM_POLICY_PROJECT,
```
yonghaoy, line 9: nit, seems we are not using IAM enum now?

google-cloudresourcemanager/src/test/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCowTest.java
```diff
@@ -47,6 +51,10 @@ public void createDeleteProject() throws Exception {
             operationCow, Duration.ofSeconds(5), Duration.ofSeconds(30));
     assertTrue(operationCow.getOperation().getDone());
     assertNull(operationCow.getOperation().getError());
+
+    Project project = managerCow.projects().get(projectId).execute();
+    assertEquals(projectId, project.getProjectId());
+
     managerCow.projects().delete(projectId).execute();
```
yonghaoy, line 58: do we need to assert project is deleted?

yonghaoy (approved):

common/src/main/java/bio/terra/cloudres/common/CloudOperation.java
wchamber: Yeah, I'm batching these so I don't have to keep bunching versions for common and its dependencies. used in next PR.

google-cloudresourcemanager/src/test/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCowTest.java
wchamber: Good point, it's not removed, but it does change its life cycle to `DELETE_REQUESTED`. Updated test, thanks.

PR #45, opened 2020-09-03
PF-30 Add get/set project IAM policy methods.

wchamber:


google-cloudresourcemanager/src/main/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCow.java
```diff
@@ -129,11 +125,65 @@ private Get(CloudResourceManager.Projects.Get get) {
...
+        this.setIamPolicy = setIamPolicy;
+      }
+
+      @Override
+      protected JsonObject serialize() {
+        return serializeProjectId(setIamPolicy.getResource());
       }
     }
+
+    private JsonObject serializeProjectId(String projectId) {
```
yonghaoy, line 182: nit: this can be static

yonghaoy (approved):

google-cloudresourcemanager/src/main/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCow.java
wchamber: this is in `class Projects` which isn't static, so it can't have static methods. We could move it out to `class CloudResourceManagerCow`, but meh.

PR #47, opened 2020-09-08
[PF-31] Add Google ServiceUsage package

wchamber:
- Include list and batchEnable methods.
- Add a testFixtures target for google-cloudresourcemanager for creating projects for tests.


google-serviceusage/src/main/java/bio/terra/cloudres/google/serviceusage/ServiceUsageOperationAdapter.java
```diff
@@ -0,0 +1,64 @@
+package bio.terra.cloudres.google.serviceusage;
+
+import bio.terra.cloudres.google.api.services.common.OperationCow;
+import com.google.api.services.serviceusage.v1.model.Operation;
+import com.google.api.services.serviceusage.v1.model.Status;
+import java.util.List;
+import java.util.Map;
+
+/** A {@link OperationCow.OperationAdapter} for {@link Operation}. */
+class ServiceUsageOperationAdapter implements OperationCow.OperationAdapter<Operation> {
```
yonghaoy, line 10: Do you think we need to extra some common things in:
https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/google-cloudresourcemanager/src/main/java/bio/terra/cloudres/google/cloudresourcemanager/ResourceManagerOperationAdapter.java

wchamber: I'm not sure what you mean. Each google api services Operation class is a different Java class, they all just look the same since they're autogenerated. This is the boilerplate to make them all look the same, i.e. look like an OperationAdapter. The boilerplate isn't interesting or particularly painful, but I'm open to suggestions. We could use reflection, but I'm not sure the marginal benefit of eliminating boilerplate is worth the risk of turning compile-time errors into runtime errors.

google-serviceusage/src/main/java/bio/terra/cloudres/google/serviceusage/ServiceUsageCow.java
```diff
@@ -0,0 +1,175 @@
...
+        JsonObject result = new JsonObject();
+        result.addProperty("parent", batchEnable.getParent());
+        result.add("content", new Gson().toJsonTree(content));
+        return result;
+      }
+    }
+
+    /** See {@link ServiceUsage.Services#list}. */
+    public List list(String parent) throws IOException {
+      return new List(services.list(parent));
```
yonghaoy, line 97: One question I have is from here:https://cloud.google.com/service-usage/docs/reference/rest/v1beta1/services/list
seems list is deprecated. But there is no document in java package:
https://developers.google.com/resources/api-libraries/documentation/serviceusage/v1/java/latest/com/google/api/services/serviceusage/v1/model/ListServicesResponse.html

Do you know what's going on there?

yonghaoy (approved):

google-serviceusage/src/main/java/bio/terra/cloudres/google/serviceusage/ServiceUsageCow.java
wchamber: Nope. It's not deprecated in v1 (instead of v1beta) which I opted to use because Data Repo was and the autogenerated javadoc links like to point to v1beta. https://cloud.google.com/service-usage/docs/reference/rest/v1/services/list
I could see us needing to update this in the future if Google Cloud ever comes back to update some of this.

PR #49, opened 2020-09-22
[PF-86] Support Get and Insert for GCP Compute subnetwork and firewall

yonghaoy:
https://broadworkbench.atlassian.net/browse/PF-86
And
https://broadworkbench.atlassian.net/browse/PF-89


google-compute/src/test/java/bio/terra/cloudres/google/compute/CloudComputeCowTest.java
```diff
@@ -59,14 +63,84 @@ public void createAndGetNetwork() throws Exception {
     assertFalse(createdNetwork.getAutoCreateSubnetworks());
   }
 
+  @Test
+  public void createAndGetSubnetwork() throws Exception {
+    Project project = createPreparedProject();
```
wchamber, line 68: optional: do we want to re-use the same project for some of these tests? If the changes we're making to it are not incompatible, that would speed them up a little.

wchamber (approved):

google-compute/src/test/java/bio/terra/cloudres/google/compute/CloudComputeCowTest.java
yonghaoy: Good point. Done

PR #54, opened 2020-10-27
[PF-204] CRL support create/get DNS zone change record 

yonghaoy:
I am not upgrading versions for othre packages because I want to have another PR to improve this

### DIFF
--- a/google-dns/src/main/java/bio/terra/cloudres/google/dns/DnsCow.java
+++ b/google-dns/src/main/java/bio/terra/cloudres/google/dns/DnsCow.java
@@ -125,28 +125,26 @@ public class DnsCow {
     public Create create(String projectId, String managedZoneName, Change change)
         throws IOException {
       return new Create(
-          changes.create(projectId, managedZoneName, change), projectId, managedZoneName, change);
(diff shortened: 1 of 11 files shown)


google-dns/src/main/java/bio/terra/cloudres/google/dns/DnsCow.java
```diff
@@ -108,4 +108,72 @@ protected JsonObject serialize() {
...
+    }
+
+    /** See {@link Dns.Changes.Create}. */
+    public class Create extends AbstractRequestCow<Change> {
+      private final String projectId;
+      private final String managedZoneName;
+      private final Change change;
+
+      public Create(
+          Dns.Changes.Create create, String projectId, String managedZoneName, Change change) {
```
wchamber, line 138: `Create` has getters for the project id and managedZoneName. Let's use those instead of duplicating the info in member variables.

wchamber (commented):
Why aren't you updating the other packages versions? I think if you don't, there can be dependency problems from the version of the cloud operations enum.

wchamber (approved):

PR #55, opened 2020-10-28
[PF-204] DNS API supports list record sets

yonghaoy:
Previous PR adding support for Change. But actuallly we are creating batch record sets, it is easier to just use list when we check existance of all record sets


google-dns/src/test/java/bio/terra/cloudres/google/dns/DnsCowTest.java
```diff
@@ -98,6 +100,10 @@ public void createAndGetChange() throws Exception {
     assertEquals("restricted.googleapis.com.", actualRecordSet.get(0).getName());
     assertEquals("199.36.153.4", actualRecordSet.get(0).getRrdatas().get(0));
     assertEquals(300, actualRecordSet.get(0).getTtl());
+
+    assertThat(
```
wchamber, line 104: also serialize test?

wchamber (approved):

PR #57, opened 2020-10-29
[no ticket] Split the central CloudOperation enum class into each package where it is used

yonghaoy:
Feels not right and painful that we have to bump all packages whenever a enum is added.

I am not updating the version here, and will use https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/56 to bump them all.


common/src/testFixtures/java/bio/terra/cloudres/testing/StubCloudOperation.java
```diff
@@ -0,0 +1,13 @@
...
+
+import bio.terra.cloudres.common.CloudOperation;
+
+/** A stub {@link CloudOperation} for test */
+public enum StubCloudOperation implements CloudOperation {
+  GOOGLE_CREATE_PROJECT {
+    @Override
+    public String getName() {
+      return "GOOGLE_CREATE_PROJECT";
+    }
```
wchamber, line 11: do we need to mimic a real operation for the tests? make a test enum value?

common/src/main/java/bio/terra/cloudres/common/CloudOperation.java
```diff
@@ -1,71 +1,7 @@
...
-  GOOGLE_GET_SUBNETWORK,
-
-  // Google DNS Operations,
-  GOOGLE_DNS_CREATE_CHANGEE,
-  GOOGLE_DNS_CREATE_ZONE,
-  GOOGLE_DNS_GET_CHANGE,
-  GOOGLE_DNS_GET_ZONE,
-  GOOGLE_DNS_LIST_RESOURCE_RECORD_SETS,
+/** The Cloud API operations */
+public interface CloudOperation {
```
wchamber, line 4: Cool. Probably worth explaining how we intend this to be used - extended with an enum class.
Also worth explaining why there's not a giant enum class.

common/src/main/java/bio/terra/cloudres/common/CloudOperation.java
```diff
@@ -1,71 +1,7 @@
...
+  /** Gets the name of operation. */
+  String getName();
```
wchamber, line 6: Can we have `name()` and get the override for free from the enum classes?

wchamber (commented):
Nice! couple nits

wchamber (approved):
thanks!

PR #58, opened 2020-11-05
big gradle normalization

zarsky-broad:
When reviewing, I recommend clicking file filter and hiding the lockfiles.


build.gradle
```diff
@@ -31,30 +28,39 @@ subprojects { subproject ->
...
 
+    configurations {
+        testImplementation.extendsFrom testFixturesImplementation
+    }
+
     dependencies {
-        api 'com.google.auth:google-auth-library-oauth2-http:0.20.0'
-        api 'com.google.cloud:google-cloud-core-http:1.93.5'
+        if (subproject.name != 'common') {
+            api project(':common')
```
wchamber, line 37: there's also the cloud-resource-schema which should not include 'common'. I'm a little worried about introducing circular dependencies here - we have to be careful that we exclude anything that 'common' depends on. I don't think the alternative of explicitly depending on common in each of the subprojects that requires 'common' is that bad.

common/src/main/java/bio/terra/cloudres/common/cleanup/CleanupRecorder.java
```diff
@@ -100,7 +99,7 @@ private static void publish(CloudResourceUid resource, ClientConfig clientConfig
     ByteString data;
     try {
       data = ByteString.copyFromUtf8(objectMapper.writeValueAsString(body));
-    } catch (JsonProcessingException e) {
+    } catch (IOException e) {
```
wchamber, line 102: why is this being changed? As is, we're explicitly wrapping the JsonProcessingException while allowing other IOExceptions to propagate. With this change, we're considering any IOException to be a serialization error, which is probably not true.
This is also a change to the 'common' package which should cause us to bump our package version.

wchamber (changes requested):
One of the "features" of how the versioning was set up in the `gradle.properties` files is that we could version the packages independently and advance them only when they changed. I don't think that's the case anymore? Any thoughts on how we can continue to do that?

common/src/main/java/bio/terra/cloudres/common/cleanup/CleanupRecorder.java
wchamber: I retract my objection, sorry, didn't look closely.
I do think this requires our version bumping, which would probably be better to do in a separate PR.

zarsky-broad: It wouldn't compile for me without this change, so I think that should probably happen first?

wchamber (approved):
thanks!

common/build.gradle
```diff
@@ -1,15 +1,17 @@
+version = '0.1.0'
```
wchamber, line 1: let's bump the version of this to 0.1.1 since we are making a code change

wchamber (approved):

PR #59, opened 2020-11-19
[PF-136] [PF-167] Support update ACL in storageCow and fix package publishing issue

yonghaoy:
setting version in subproject's gradle doesn't work, parent package is not able get the value. Switch to use `gradle.proprites`.

Also change log level to info. I did them in the same PR because/bumping version and publish package takes few rounds, and logger is a small change. Let me know if I should split this


common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -140,7 +137,7 @@ void logEvent(
 
     logEntry.add("request", request);
 
-    logger.debug(logEntry.toString());
+    logger.info(logEntry.toString());
```
wchamber, line 140: hmmm. I'm not totally sold that we should log info every cloud operation. Could be persuaded. Maye let's just try it and see how it goes.
Another option is to keep this at debug, but logger.info whenever there's an executionException. I think that's often when we're trying to see what's going on.

google-storage/src/test/java/bio/terra/cloudres/google/storage/StorageIntegrationUtils.java
```diff
@@ -44,4 +50,16 @@ static BlobCow createBlobWithContents(StorageCow storageCow, BlobId blobId, Stri
     }
     return storageCow.get(blobId);
   }
+
+  /** Helper to compare two list of {@link Acl}'s entity and role, but ignores the etag and id. */
+  static void assertAclsMatch(List<Acl> expect, List<Acl> actual) {
```
wchamber, line 55: mind double checking if we have this logic already in one of the tests? If so, use this function there. I'm not sure, but feels very familiar.

wchamber (approved):

google-storage/src/test/java/bio/terra/cloudres/google/storage/StorageIntegrationUtils.java
yonghaoy: we have a function compare one Acl, not the same

common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: changed to use info when having exception, use debug if not.

PR #60, opened 2020-11-19
[PF-229]: support delete firewall

yonghaoy:
Version update will be done batch in sperate PR as I have few other upcoming changes


google-compute/src/main/java/bio/terra/cloudres/google/compute/CloudComputeCow.java
```diff
@@ -209,7 +209,7 @@ public Insert(Compute.Firewalls.Insert insert, String projectId, Firewall firewa
         this.firewall = firewall;
         this.projectId = projectId;
       }
-
+
```
wchamber, line 212: odd whitespace. spotlessApply?

wchamber (approved):

PR #61, opened 2020-11-19
[no ticket] make blob & bucket not public in test

yonghaoy:
It is recommanded from Broad InfoSecs that not to make cloud resource public.


wchamber (approved):
Worth a description in the PR why we're doing this.

PR #63, opened 2020-11-23
[PF-238] CRL support create/list/delete GCP service account

yonghaoy:
Should be the same structure as other APIs.


google-iam/src/main/java/bio/terra/cloudres/google/iam/IamCow.java
```diff
@@ -0,0 +1,142 @@
...
+        return new Create(serviceAccounts.create(name, content), name, content);
+      }
+
+      public class Create extends AbstractRequestCow<ServiceAccount> {
+        private final String name;
+        private final CreateServiceAccountRequest content;
+
+        public Create(
+            Iam.Projects.ServiceAccounts.Create create,
+            String name,
```
gjuggler, line 81: Parameter naming: there's not much guidance for a developer to understand the requirements around this "name". I had to cross-reference against the integration test to understand that this is actually a GCP project identifier in the form "projects/projectId". I think this is technically called the "relative resource name" (per [these docs](https://cloud.google.com/apis/design/resource_names#relative_resource_name)), though that doesn't exactly roll off the tongue.

Aha – we can just use the terminology / guidance from Google's API docs [here](https://developers.google.com/resources/api-libraries/documentation/iam/v1/java/latest/com/google/api/services/iam/v1/Iam.Projects.ServiceAccounts.html#create-java.lang.String-com.google.api.services.iam.v1.model.CreateServiceAccountRequest-).

Alternatively, you could make this constructor take a projectId only, and do the templating of "project/{projectId}" itself. That would be more consistent with our use of "projectId" elsewhere, though I'm not sure we want to set a pattern where the Cow objects are doing string manipulation like that.

(Whatever solution you decide on, it should apply to the List & Delete constructors as well, since they have a similar "name" param.)

google-iam/src/test/java/bio/terra/cloudres/google/iam/IamCowTest.java
```diff
@@ -0,0 +1,114 @@
...
+        IntegrationUtils.DEFAULT_CLIENT_CONFIG,
+        IntegrationCredentials.getAdminGoogleCredentialsOrDie());
+  }
+
+  @Test
+  public void createAndListAndDeleteServiceAccount() throws Exception {
+    IamCow iam = defaultIam();
+    Project project = createPreparedProject();
+    String projectId = project.getProjectId();
+    String projectName = "projects/" + projectId;
```
gjuggler, line 38: terminology nit: call this variable "resourceName," since "project name" is a different concept representing the human-readable & mutable name given to a project.

google-iam/src/test/java/bio/terra/cloudres/google/iam/IamCowTest.java
```diff
@@ -0,0 +1,114 @@
...
+    String accountId = randomSserviceAccountIdName();
+    ServiceAccount serviceAccount =
+        iam.projects()
+            .serviceAccounts()
+            .create(
+                "projects/" + projectId, new CreateServiceAccountRequest().setAccountId(accountId))
```
gjuggler, line 44: nit: replace "projects/" + projectId with projectName variable

gjuggler (approved):

google-iam/src/main/java/bio/terra/cloudres/google/iam/IamCow.java
yonghaoy: For all those GCP APIS, we find Google's API doc already has detailed explaination. So we have comment like
`/** See {@link Iam.Projects.ServiceAccounts#create(String, CreateServiceAccountRequest)}. */` for all CRL methods. So we didn't add any our own javadoc as it is redundant. Developer should be able to jump to Google's API doc easily from their. IDE and understand the parameters, and input, example.

That is kind of CRL pattern

gjuggler: Ah, gotcha, I'd missed that it gets quick-linked by the IDE. That's great 👍

PR #65, opened 2021-02-10
[PF-412] Add custom role endpoint support

zloery:
This change adds wrappers around Google's [custom role management endpoints](https://cloud.google.com/iam/docs/reference/rest/v1/projects.roles) to the IAM CRL to support creating, deleting, patching, getting, and deleting custom roles on a project-level.

This does not wrap organization-level custom role endpoints. Those are similar, but we do not currently have a usecase for them.


google-iam/src/main/java/bio/terra/cloudres/google/iam/IamCow.java
```diff
@@ -138,5 +138,142 @@ protected JsonObject serialize() {
...
+         */
+        public List setView(String view) {
+          setField(VIEW_FIELD_NAME, view);
+          return this;
+        }
+
+        @Override
+        protected JsonObject serialize() {
+          JsonObject result = new JsonObject();
+          result.addProperty("parent", parent);
```
yonghaoy, line 249: this serialize is used to log entire request.  I think it would be useful if we can log view if it has it.  
Maybe make an Optional<String> for that, and serialize it if present?

yonghaoy (approved):
you will also need to update version here, https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/google-iam/gradle.properties
Then manually publish library after merge

google-iam/src/test/java/bio/terra/cloudres/google/iam/IamCowTest.java
```diff
@@ -97,6 +100,108 @@ public void deleteServiceAccountSerialize() throws Exception {
...
+            .execute();
+    // Retry 6 times to make sure get after create works.
+    int retryNum = 0;
+    List<Role> listResult = null;
+    while (retryNum < 6) {
+      listResult = iam.projects().roles().list(resourceName).execute().getRoles();
+      if (listResult != null) {
+        break;
+      }
+      retryNum++;
```
wchamber, line 125: nit: use a for loop for retryNum?
I'm sad this takes up to 18seconds.

google-iam/src/test/java/bio/terra/cloudres/google/iam/IamCowTest.java
```diff
@@ -118,4 +223,13 @@ public static String randomServiceAccountIdName() {
   private static String fullServiceAccountName(String projectId, String accountId) {
     return String.format("projects/%s/serviceAccounts/%s", projectId, accountId);
   }
+
+  /**
+   * Create a Role object with the permission iam.roles.create and no other fields specified.
+   *
+   * @return
```
wchamber, line 230: nit: remove the empty `@return` block?

wchamber (changes requested):

google-iam/src/main/java/bio/terra/cloudres/google/iam/IamCow.java
wchamber: there exists List#getView  https://developers.google.com/resources/api-libraries/documentation/iam/v1/java/latest/com/google/api/services/iam/v1/Iam.Projects.Roles.List.html#getView--

zloery: Restructured this a bit, view is now a field and will be serialized

google-iam/src/main/java/bio/terra/cloudres/google/iam/IamCow.java
```diff
@@ -138,5 +138,142 @@ protected JsonObject serialize() {
...
+
+        /**
+         * Optional view for the returned Role objects. When `FULL` is specified, the
+         * `includedPermissions` field is returned, which includes a list of all permissions in the
+         * role. The default value is `BASIC`, which does not return the `includedPermissions`
+         * field.
+         */
+        public List setView(String view) {
+          setField(VIEW_FIELD_NAME, view);
+          return this;
```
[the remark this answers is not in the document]
zloery: This is mostly an issue of the underlying `Iam.Projects.Roles.List` object not being exposed. I switched to making that visible to `AbstractRequestCow`s, let me know if there's a better way to structure this

google-iam/src/test/java/bio/terra/cloudres/google/iam/IamCowTest.java
zloery: I haven't actually timed it, this was mainly for consistency with above. I'm not sure what the expected duration is, other than definitely not 18s

google-iam/src/main/java/bio/terra/cloudres/google/iam/IamCow.java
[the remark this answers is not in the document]
wchamber: Sorry, there's a pattern for this. Basically, it's cows all the way down. While we don't want to expose the underlying List object through our wrapping Cow, it's expected and correct to wrap its methods. In this case, our List COW takes the underlying List in it's constructor. We should store the underlying list as a `private final Iam.Projects.Roles.List list;` here in our List COW class. Then, our COW List setView is 
```
public List setView(String view) {
          list.setView(view);
          return this;
}
```
Similarly, in the serialize we can use `list.getView()`.
Example: https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/de4de2e29cc62330b49cceb2a1976eb2c452caeb/google-serviceusage/src/main/java/bio/terra/cloudres/google/serviceusage/ServiceUsageCow.java#L100-L127

wchamber (changes requested):

google-iam/src/main/java/bio/terra/cloudres/google/iam/IamCow.java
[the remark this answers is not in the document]
zloery: Ah perfect, I was looking for how other cows did this but didn't see anything. That approach makes sense to me

wchamber (approved):
thanks!

PR #66, opened 2021-02-25
[PF-506] change version

jaycarlton:


wchamber (commented):
You also need to update the gradle lockfile. https://github.com/DataBiosphere/terra-cloud-resource-lib#dependencies

jaycarlton:
I'm seeing AbstractMethodError failures in the tests. Maybe some assumption about the library doesn't hold anymore in the new version. I'll dig into this a bit later.

wchamber:
> I'm seeing AbstractMethodError failures in the tests. Maybe some assumption about the library doesn't hold anymore in the new version. I'll dig into this a bit later.

fwiw, I'd guess we'd need to update more dependencies. may be some breaking change in underlying shared google libraries.

PR #67, opened 2021-03-02
[PF-231] Support set and get Bucket IAM Policy in StorageCow

yonghaoy:
Background: RBS needs this to setup storage log bucket. Right now RBS uses. legacy ACL, and that does not work in Verily cloud. 
https://cloud.google.com/storage/docs/access-logs

Getting Read timed out error for some DnsCow test. Also increase the default timeout to fix it.


yonghaoy:
@dionloy FYI

common/src/testFixtures/java/bio/terra/cloudres/testing/IntegrationUtils.java
```diff
@@ -37,4 +38,17 @@ public static String randomName() {
   public static String randomNameWithUnderscore() {
     return UUID.randomUUID().toString().replace('-', '_');
   }
+
+  /**
+   * Sets longer timeout because some operation(e.g. Dns.ManagedZones.Create) may take longer than
+   * default timeout.
```
gjuggler, line 44: nit: please mention the purpose of the requestInitializer parameter, either with a @param tag or in the method description. Something like "Accepts a requestInitializer to allow chaining, since API clients have exactly one initializer and credentials are typically required as well."

gjuggler (approved):

PR #69, opened 2021-03-10
PF-468 Add GoogleAiNotebookUid to the cloud resources uids.

wchamber:
Adding this first to update the Janitor and its client library to use in clean up when creating AI notebook instances.

### DIFF
--- a/cloud-resource-schema/src/main/resources/cloud_resources_uid.yaml
+++ b/cloud-resource-schema/src/main/resources/cloud_resources_uid.yaml
@@ -10,7 +10,7 @@ paths: {}
 
 components:
   schemas:
-    GoogleAiNotebookUid:
+    GoogleAiNotebookInstanceUid:
       type: object
       required:
         - projectId
@@ -21,8 +21,10 @@ components:
           type: string
         location:
           type: string
+          description: The location of the instance, e.g. 'uswest1-b'.
         instanceId:
           type: string
+          description: The user-provided instance id to identify the instance.
     GoogleBigQueryDatasetUid:
       type: object
       required:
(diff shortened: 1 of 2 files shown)


cloud-resource-schema/src/main/resources/cloud_resources_uid.yaml
```diff
@@ -10,6 +10,19 @@ paths: {}
...
+      required:
+        - projectId
+        - location
+        - instanceId
+      properties:
+        projectId:
+          type: string
+        location:
+          type: string
+        instanceId:
```
jaycarlton, line 24: maybe add descriptions to these. I'm not familiar with the instance ID format.

jaycarlton (approved):

cloud-resource-schema/src/main/resources/cloud_resources_uid.yaml
wchamber: took a stab, ptal. Like most things in CRL, needs some knowledge of the underlying objects https://cloud.google.com/ai-platform/notebooks/docs/reference/rest/v1/projects.locations.instances

yonghaoy (approved):

yonghaoy (commented):
And also bump schema version

cloud-resource-schema/src/main/resources/cloud_resources_uid.yaml
jaycarlton: thanks

wchamber:
> And also bump schema version

sync'd offline, that's the `cloud-resource-schema/gradle.properties` which we are updating.

PR #71, opened 2021-03-15
[PF-539] Add structured logs to the CRL operation annotator

gjuggler:
This PR refactors the OperationAnnotator in a few ways. The most important change is a switch from appending string-formatted JSON in the log message to providing a separate JsonObject argument to the logger call. The intent here is for the JsonObject to be picked up by the Terra common logging package (for services where it's in the classpath), and included in any JSON-formatted output logs.

The other main change is to include the operation duration in the structured and string-formatted output. This helps give a quick view on whether a given operation was quick or long-lived.

**Testing**
I manually tested this locally by integrating with RBS. With human-readable logging turned on, the simple string-formatted log message was shown. With JSON logging turned on, structured data showed up in the JSON output as expected.


common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -112,39 +109,60 @@ private void recordLatency(Duration duration, CloudOperation operation) {
...
+          // Include logData for terra-common-lib logging to pick up and include in JSON output.
+          logData);
     }
   }
 
+  // Turns the default Duration.toString output into a more human-readable output. See
+  // https://stackoverflow.com/a/40487511 for inspiration.
+  private String prettyPrintDuration(Duration duration) {
+    // Truncate to 0.1 second precision.
+    duration = Duration.ofMillis((duration.toMillis() / 100l) * 100l);
```
jaycarlton, line 162: make a new variable here instead of overwriting the parameter.

common/src/test/java/bio/terra/cloudres/common/cleanup/CleanupRecorderTest.java
```diff
@@ -29,7 +29,7 @@
 import org.mockito.ArgumentCaptor;
 import org.mockito.Mock;
 
-@Tag("unit")
+@Tag("integration")
```
gjuggler, line 32: I think this was a mis-labeled test. It uses integration credentials and was failing locally for me when running unit tests.

common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -112,39 +109,60 @@ private void recordLatency(Duration duration, CloudOperation operation) {
...
+    return duration.toString().substring(2).replaceAll("(\\d[HMS])(?!$)", "$1 ").toLowerCase();
```
jaycarlton, line 163: I'm surprised there's not a way to do this with a format string. A comment about what the output format is might help.

common/src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
```diff
@@ -24,16 +29,14 @@
...
   private static final String CLIENT = "TestClient";
-  private static final String PROJECT_INFO_STRING =
-      "{\"name\":\"myProj\",\"projectId\":\"project-id\",\"labels\":{\"k1\":\"v1\",\"k2\":\"v2\"}}";
-  private static final JsonObject PROJECT_INFO_JSON_OBJECT =
-      new Gson().fromJson(PROJECT_INFO_STRING, JsonObject.class);
-  private static final String TRACE_ID = "1234567890123456";
+  private static final Gson JSON_PARSER = new GsonBuilder().setLenient().create();
+  // A fake but somewhat-realistic create-project request payload
+  private static final JsonObject PROJECT_REQUEST =
+      JSON_PARSER.fromJson(
```
jaycarlton, line 35: so the lenient mode lets it accept single quotes?

common/src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
```diff
@@ -56,10 +59,12 @@
 
   private static final OperationAnnotator.CowSerialize SERIALIZE =
       () -> {
-        return PROJECT_INFO_JSON_OBJECT;
+        return PROJECT_REQUEST;
       };
 
-  ArgumentCaptor<String> logArgument = ArgumentCaptor.forClass(String.class);
+  ArgumentCaptor<String> stringArgumentCaptor = ArgumentCaptor.forClass(String.class);
```
jaycarlton, line 65: you can also use `@ArgumentCaptor`

common/src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
```diff
@@ -128,78 +133,51 @@ public void testExecuteGoogleCloudCall_withCheckedException() throws Exception {
...
 
-    // Expected result in Json format
-    verify(mockLogger).debug(logArgument.capture());
-    assertEquals(
-        EXPECTED_LOG_PREFIX + "\"request\":" + PROJECT_INFO_STRING + "}", logArgument.getValue());
+    verify(mockLogger).debug(stringArgumentCaptor.capture(), gsonArgumentCaptor.capture());
+    JsonObject json = gsonArgumentCaptor.getValue();
+    assertThat(
+        stringArgumentCaptor.getValue(),
+        Matchers.containsString("CRL completed TEST_OPERATION (2.3s"));
```
jaycarlton, line 150: should there be a closing paren?

common/src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
```diff
@@ -128,78 +133,51 @@ public void testExecuteGoogleCloudCall_withCheckedException() throws Exception {
...
+    assertThat(json.getAsJsonPrimitive("clientName").getAsString(), Matchers.equalTo("TestClient"));
+    assertThat(json.getAsJsonPrimitive("durationMs").getAsLong(), Matchers.equalTo(2345l));
```
jaycarlton, line 152: nit: we generally statically import the matchers, since they're obvious in context. Also `assertEquals` is easier and cleaner IMO.

common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -82,16 +81,14 @@ public OperationAnnotator(ClientConfig clientConfig, Logger logger) {
       tracer
           .getCurrentSpan()
           .putAttribute(
-              "httpErrorCode", AttributeValue.longAttributeValue(httpErrorCode.orElse(-1)));
+              "/terra/crl/httpErrorCode",
```
jaycarlton, line 84: Why does this attribute need a prefix when the JSON properties don't?

jaycarlton (commented):
Can you please include screenshots from Cloud Logging and/or the terminal.

common/src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
gjuggler: Ack on both counts – I was adapting to the local code context where static imports hadn't been done, but I agree there's no reason not to. Updated.

common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: that is the key of this span's attribute.

common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -112,39 +110,65 @@ private void recordLatency(Duration duration, CloudOperation operation) {
...
 
+  // Turns the default Duration.toString output into a more human-readable output (e.g. "4h 2m 3s").
+  // See https://stackoverflow.com/a/40487511 for inspiration.
+  private String prettyPrintDuration(Duration duration) {
+    return Duration
+        // Truncate to 0.1 second precision.
+        .ofMillis((duration.toMillis() / 100l) * 100l)
+        // Format the duration and apply some fine-tuning to the string output.
+        .toString()
+        .substring(2)
```
yonghaoy, line 167: nit: add:   ` // Remove the PT`

common/src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
gjuggler: Thanks for the tip – I decided not to change for now, given the pre-existing pattern.

Are there good examples of its use in AoU or Terra that I might take a look at?

common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -112,39 +110,65 @@ private void recordLatency(Duration duration, CloudOperation operation) {
...
+    logData.addProperty("operation", operation.name());
+    logData.add("requestData", requestData);
 
-    logEntry.add("request", request);
-
-    // Log as debug level if no exceptions, otherwise use info level.
     if (executionException.isPresent()) {
-      logger.info(logEntry.toString());
+      OptionalInt httpErrorCode = getHttpErrorCode(executionException.get());
+      logger.error(
```
yonghaoy, line 143: Bumping from into to error might be too aggressive? e.g., WSM and Buffer makes a get call before creating project, exception is expected in that cases. So CRL just logs as `info` for exceptions, `debug` for all others. 
 CRL's clients catch CRL exception then logs themselves if they think it should be an error or warning.

common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -112,39 +110,65 @@ private void recordLatency(Duration duration, CloudOperation operation) {
...
+          String.format(
+              "CRL exception in %s (HTTP code %s, %s)",
```
yonghaoy, line 145: In CRL, it used to be a `String plus Json.` But we want Stackdriver can parse this log message, so we make the whole log message be completely structured. But that time, I am not sure how Stackdriver parse logs. This seems to be a better format, if Stackdriver can understand string + json

common/src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
jaycarlton: Yeah, it's mainly in AoU. I don't know if the usages I added in WSM are still there.

common/src/test/java/bio/terra/cloudres/common/cleanup/CleanupRecorderTest.java
yonghaoy: This does not make GCP calls, and mock pubsub, everything is in local, not connected. So we make this 'unit'. It might be hard(or I am just lazy do to so) to mock janitor credential creation, so we just let the test to load credentials but not actually use it.
I am fine with `Integration` or `Unit`

yonghaoy (approved):

yonghaoy (approved):

common/src/test/java/bio/terra/cloudres/common/cleanup/CleanupRecorderTest.java
gjuggler: Helpful context! I'll still lean towards `Integration`, since it requires some level of credential gathering to have happened. My instinct for `unit` is that someone should be able to get them to pass without any access to Vault or cloud creds.

I'll add a comment trying to capture a bit of this nuance for posterity.

gjuggler (commented):
> this was `info` before the change

Ah – sorry, I accidentally pushed some commits without corresponding comments. Had meant to hold both back until tomorrow given the hour, but accidentally pushed the commits too soon. See my other comment – I think debug is best for consistency here.

common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
yonghaoy: Debug level here but Buffer have debugging level turn on SG!

jaycarlton:
https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/71#discussion_r594823392

That's the agent we weren't able to use in AoU because of GAE. It's coming back to me.

gjuggler:
Side note: I had a lot of trouble running CI tests for this PR! It took several retries to get a clean run. I was seeing 502 bad gateway errors when Gradle was loading packages, then some flakiness in the `MetricsHelperTest`. I filed https://broadworkbench.atlassian.net/browse/PF-571 to track the issues.

jaycarlton (approved):

PR #73, opened 2021-03-22
PF-468 Add initial google-notebooks package

wchamber:
add an initial Google AI Notebooks package with create/get/delete for instances and get operation.
Update the janitor client for the new version that includes the AI Notebooks.
Refactor the CloudBillingUtils test fixture so that each test does not duplicate the common billing package.


wchamber:
Tests currently breaking because of Java 8 vs 11 mismatch on Janitor client, but otherwise ready for review.

google-notebooks/src/main/java/bio/terra/cloudres/google/notebooks/AIPlatformNotebooksCow.java
```diff
@@ -0,0 +1,247 @@
...
+
+    /** See {@link AIPlatformNotebooks.Projects.Locations.Instances#create(String, Instance)}. */
+    public Create create(String parent, Instance instance) throws IOException {
+      return new Create(instances.create(parent, instance), instance);
+    }
+
+    /** Create an instance with the {@link InstanceName}. See {@link #create(String, Instance)}. */
+    public Create create(InstanceName instanceName, Instance instance) throws IOException {
+      Preconditions.checkArgument(
+          instance.getName() == null || instanceName.formatName().equals(instance.getName()),
```
yonghaoy, line 74: `instance.getName() == null `means here?

yonghaoy (approved):

google-notebooks/src/main/java/bio/terra/cloudres/google/notebooks/AIPlatformNotebooksCow.java
wchamber: The `Instance` model object describes how GCP should create the instance. The `name` field does not need to be specified on the `Instance` object passed into creation: the name can be set on the `Create` object. But if it is set on the `Instance` object and as a part of the `InstanceName` and they differ, the user's going to have a bad time as only one of the names will actually be used. so `== null || instanceName.formatName().equals(instance.getName)` checks that it's unspecified or equivalent. Tried modifiying the comment.
If the `Instance` model object is retrieved from GCP, it will always have the name set.

zloery (approved):
I'm not super familiar with the larger gradle setup, but generally LGTM.

If we're planning to add notebooks as controlled resources in WSM, in the future we'll also need wrappers for `getIamPolicy` and `setIamPolicy`. I'm fine expanding the set of endpoints as things get clearer in the future, though

wchamber:
> I'm not super familiar with the larger gradle setup, but generally LGTM.
> 
> If we're planning to add notebooks as controlled resources in WSM, in the future we'll also need wrappers for `getIamPolicy` and `setIamPolicy`. I'm fine expanding the set of endpoints as things get clearer in the future, though

yep, plan on adding those and a list command in a future PR. I think I made this long enough for both of you :)

google-notebooks/gradle.properties
```diff
@@ -0,0 +1 @@
+version = 0.0.0
```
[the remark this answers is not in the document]
wchamber: sgtm

wchamber:
@rtitle retroactive fyi

PR #74, opened 2021-03-23
PF-468 Add a list notebooks instance operation.

wchamber:

PR #75, opened 2021-03-23
PF-468 Add get and set policy operations for AI Notebook Instances

wchamber:


wchamber:
@rtitle fyi more CRL operations for AI Notebook instances

zloery (approved):
don't forget to bump gradle.properties if needed, probably depends on the ordering with #74.

yonghaoy (approved):

wchamber:
> don't forget to bump gradle.properties if needed, probably depends on the ordering with #74.
fwiw, the version bumping's moot until we publish a version of this library, which is a manual process today.

PR #79, opened 2021-04-19
[PF-697] Move bigquery to wrap generated google API

zloery:
The google-bigquery project currently wraps the `com.google.cloud:google-cloud-bigquery` client library, but this does not support the full set of dataset-level IAM operations available in the REST api which WSM needs. 

Google provides another client library generated from available APIs at `com.google.apis:google-api-services-bigquery`, and this change moves the google-bigquery CRL project to wrap that client instead.

This is a breaking change and would normally be a major version bump, but given that we're still on major version 0 I'm inclined to just increment the minor version.


zloery:
Tests are failing on GHA as the actions use Java 8 instead of 11, but pass locally. Broke that change out into a separate PR (#80) so it doesn't get lost in all the changes here.

google-bigquery/src/test/java/bio/terra/cloudres/google/bigquery/ResourceTracker.java
```diff
@@ -3,47 +3,83 @@
...
+import com.google.api.services.bigquery.model.DatasetReference;
+import com.google.api.services.bigquery.model.Table;
+import com.google.api.services.bigquery.model.TableReference;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import javax.annotation.Nullable;
 
 /** Helper class to track BigQuery resources created in tests and do best effort clean up. */
 public class ResourceTracker {
```
wchamber, line 16: This predates Janitor and we should probably just delete it.

google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
```diff
@@ -1,206 +1,646 @@
...
+      @Override
+      protected JsonObject serialize() {
+        JsonObject result = new JsonObject();
+        result.addProperty("projectId", getProjectId());
+        result.addProperty("datasetId", getDatasetId());
+        return result;
+      }
+    }
+
+    public Insert insert(String projectId, Dataset content) throws IOException {
```
wchamber, line 145: javadoc. Ditto other methods on the cow objects

google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
```diff
@@ -1,206 +1,646 @@
...
+      CloudResourceUid datasetUid =
+          new CloudResourceUid()
+              .googleBigQueryDatasetUid(
+                  new GoogleBigQueryDatasetUid()
+                      .projectId(projectId)
+                      .datasetId(content.getDatasetReference().getDatasetId()));
+      CleanupRecorder.record(datasetUid, clientConfig);
```
wchamber, line 152: subclasses of AbstractRequestCow should override `resourceUidCreation` for cleanup recording.
https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/1d2d0726960248f82af7b490f34e44cdd9dc3246/google-api-services-common/src/main/java/bio/terra/cloudres/google/api/services/common/AbstractRequestCow.java#L40-L49

google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
```diff
@@ -1,206 +1,646 @@
...
+        JsonObject result = new JsonObject();
+        result.addProperty("projectId", getProjectId());
+        result.addProperty("datasetId", getDatasetId());
+        result.addProperty("tableId", getTableId());
+        return result;
+      }
+    }
+
+    // Unlike every other table endpoint, the IAM APIs take a single resource identifier string
+    // instead of a projectId + datasetId + tableId. I've hidden this quirk for consistency.
```
wchamber, line 374: another option here is to add a value class for the (projectId, datasetId, tableId) tuple. I started doing that for the AI Notebook instances and have been happy with how it turned out. It also provides a nice place to have the functions to parse the resource name tuple.
That's probably more helpful the more APIs are using the format.
https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/google-notebooks/src/main/java/bio/terra/cloudres/google/notebooks/InstanceName.java

OTOH, it looks like the bigquery lib provides 'TableReference' which does most of what we'd want, other than parsing (and possible equality comparison?) so maybe use that.

google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
```diff
@@ -1,206 +1,646 @@
...
+
+    public Insert insert(String projectId, String datasetId, Table content) throws IOException {
+      CloudResourceUid tableUid =
+          new CloudResourceUid()
+              .googleBigQueryTableUid(
+                  new GoogleBigQueryTableUid()
+                      .projectId(projectId)
+                      .datasetId(datasetId)
+                      .tableId(content.getTableReference().getTableId()));
+      CleanupRecorder.record(tableUid, clientConfig);
```
wchamber, line 418: ditto override `resourceUidCreation`

google-bigquery/src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
```diff
@@ -28,210 +49,386 @@
...
+        .tables()
+        .insert(datasetReference.getProjectId(), datasetReference.getDatasetId(), tableToCreate)
+        .execute();
+
+    Dataset fetchedDataset =
+        bigQueryCow.datasets().get(projectId, datasetReference.getDatasetId()).execute();
+    assertNotNull(fetchedDataset);
+
+    // Attempt a delete call with deleteContents flag set to false. Because this dataset contains a
+    // table, this should fail.
```
wchamber, line 113: I don't think we need to be testing that Google's APIs work as advertised. These tests are just to smoke test that we can call all of them.

yonghaoy (commented):
What's the plan for existing CRL client to migrate?

google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
zloery: ah, I hadn't noticed that one. thanks!

google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
zloery: Yeah - I'm a little hesitant to use a TableReference as the parameter when most underlying methods are really taking the projectId + datasetId + tableId. I think it would be more useful if more than these three IAM methods used a single String resource argument, but given the rest of the dataset/table interface I'd prefer to use the individual argument interface here.

google-bigquery/src/test/java/bio/terra/cloudres/google/bigquery/BigQueryCowTest.java
zloery: sgtm. Other delete calls exercise `setDeleteContents`, so that should validate that it can be set

google-bigquery/build.gradle
```diff
@@ -1,6 +1,14 @@
...
+    api project(':google-api-services-common')
     implementation project(':cloud-resource-schema')
 
-    api group: 'com.google.cloud', name: 'google-cloud-bigquery', version: '1.116.0'
+    api group: 'com.google.apis', name: 'google-api-services-bigquery', version: 'v2-rev20201030-1.30.10'
+
+    testFixturesImplementation(testFixtures(project(':common')))
+    testImplementation project(':google-cloudresourcemanager')
+    testImplementation(testFixtures(project(':google-api-services-common')))
+    testImplementation(testFixtures(project(':google-billing')))
```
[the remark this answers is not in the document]
zloery: ah, I tried removing these earlier and got an error - turns out that was a lockfile issue, not a real dependency.

google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
zloery: > What's the plan for existing CRL client to migrate?

AFAIK the only clients are WSM and janitor - I'm planning to publish this change as a new version, then migrate both clients immediately after

zloery:
> What's the plan for existing CRL client to migrate?

AFAIK the only clients are WSM and Janitor. I'm planning to publish this change as a new version, then update both clients to use it immediately

google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
```diff
@@ -1,206 +1,678 @@
...
+        result.addProperty("datasetId", getDatasetId());
+        result.addProperty("tableId", getTableId());
+        result.add("content", new Gson().toJsonTree(update.getJsonContent()).getAsJsonObject());
+        return result;
+      }
+    }
+
+
+
+    private String tableResourceName(String projectId, String datasetId, String tableId) {
```
wchamber, line 674: consider making this public (w/ javadoc) since it's some boring formatting that could be needed for other parts of the api.

google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java
wchamber: Not totally following, I think the TableReference has project, dataset, and table ids. https://developers.google.com/resources/api-libraries/documentation/bigquery/v2/java/latest/com/google/api/services/bigquery/model/TableReference.html?is-external=true
Fine to leave as is, but I think making a little auto value would prove helpful for clients, and making it harder to accidentally switch the 3 string arguments.

wchamber (approved):

yonghaoy (approved):

PR #80, opened 2021-04-19
[No ticket] Update java version from 8 to 11

zloery:
This change moves all of CRL from Java 8 to Java 11. All sub-project version numbers were incremented for this, and I believe all CRL clients are already using Java 11, so this should not be a disruptive change.


yonghaoy (commented):
Do we need to upgrade java version in `build.gradle`? 
https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/build.gradle#L15

zloery:
> Do we need to upgrade java version in `build.gradle`?
> https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/build.gradle#L15

Actually, I'm not quite sure. Other repos moving to Java 11 have done that, but I wonder if there are compatibility concerns if CRL clients are still on Java 8. I also don't know the full list of CRL clients or if any are still on 8, but I can look into that.

yonghaoy:
> > Do we need to upgrade java version in `build.gradle`?
> > https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/build.gradle#L15
> 
> Actually, I'm not quite sure. Other repos moving to Java 11 have done that, but I wonder if there are compatibility concerns if CRL clients are still on Java 8. I also don't know the full list of CRL clients or if any are still on 8, but I can look into that.

All CRL clients are one java 11. It was in Java 8 because Data Repo was in Java 8, but now I think we can upgrade CRL to 11. You can just upgrade the `build.gradle` in the same PR I think.

wchamber (changes requested):
Please also update the `.java-version` file and change the PR description to say we're moving all of CRL to java 11.

We should probably update all the `gradle.properties` version values in all the packages a minor increment.
(It'd be nice if we set this up to be an automated process, but afaik we're not sure how to do this with multiple packages without bumping all their versions for an unrelated package change.)

wchamber (approved):

yonghaoy (approved):

PR #85, opened 2021-05-17
PF-744 Normalize versions with Stairway and TCL

ddietterich:
In addition to version changes, I had to switch to Gson from Jackson in Default.java since the Jackson version has been deprecated and removed from guava.


build.gradle
```diff
@@ -35,20 +35,28 @@ subprojects { subproject ->
         ext {
             janitorclient = '0.3.0-SNAPSHOT'
         }
-        api group: 'com.google.auth', name: 'google-auth-library-oauth2-http', version: '0.20.0'
-        api group: 'com.google.cloud', name: 'google-cloud-core-http', version: '1.93.5'
+        // Google dependencies - make sure we are using -jre and not android and use common bom
+        constraints {
+            implementation 'com.google.guava:guava:30.1.1-jre' // "-jre" for Java 8 or higher
+            testFixturesImplementation 'com.google.guava:guava:30.1.1-jre' // "-jre" for Java 8 or higher
```
wchamber, line 41: How does this compare to `plugins { id("de.jjohannes.missing-metadata-guava") version "0.4" }` per https://blog.gradle.org/guava?
I'm happy to defer to others opinions here. @yonghaoy were you looking at some of this?

google-api-services-common/src/main/java/bio/terra/cloudres/google/api/services/common/Defaults.java
```diff
@@ -3,10 +3,12 @@
 import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
 import com.google.api.client.http.HttpTransport;
 import com.google.api.client.json.JsonFactory;
-import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.client.json.gson.GsonFactory;
 import java.io.IOException;
 import java.security.GeneralSecurityException;
 
+// import com.google.api.client.json.jackson.JacksonFactory;
```
wchamber, line 10: remove me?

wchamber (approved):

build.gradle
ddietterich: I got the recipe I used from [here](https://github.com/GoogleCloudPlatform/cloud-opensource-java/wiki/The-Google-Cloud-Platform-Libraries-BOM#guava-versions--jre-or--android)

jmczerk: If we can get the desired behavior using a platform rather than a plugin, I'd lean towards that.  Ideally I'd like to get to a point where we can consistently use platforms to manage dependencies and avoid disparate plugins (still in process of determining if this is achievable).  That said, I'd be interested in hearing @yonghaoy inputs.

jmczerk (approved):

wchamber:
Oops, I forgot we should have bumped all of the packages properties in this PR. We don't have automatic version updating for our multi package repo.
I expect the automatic release to fail with artifactory permission error, per PF-722
I can send a PR and shepherd a new CRL release.

ddietterich:
Thank you!

PR #86, opened 2021-05-17
[PF-417] add missing option

jaycarlton:
Add an option to update the storage class on the underlying GCS API object.


wchamber (approved):

wchamber:
Heads up we're going to race on the gradles.properties modification, see #87.
We'll also need to manually release the new version of this per PF-722. If you like, I'll do it if you ping me once this is merged.

zloery (approved):

PR #87, opened 2021-05-17
PF-744 Bump all the gradle.properties for PR #85.

wchamber:
I forgot to check for this in #85. We haven't figured out a way to automate this on a per-package level yet, so we're still at manually modifying the `gradle.properties` files used to create the release version, which at least lets us increment the packages independently.

### DIFF
--- a/google-storage/gradle.properties
+++ b/google-storage/gradle.properties
@@ -1,1 +1,1 @@
-version = 0.8.0
+version = 0.9.0
--- a/google-storage/gradle.properties
+++ b/google-storage/gradle.properties
@@ -1,1 +1,1 @@
-version = 0.9.0
+version = 0.10.0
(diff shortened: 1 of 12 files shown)


jmczerk (approved):

google-storage/gradle.properties
```diff
@@ -1 +1 @@
-version = 0.9.0
```
jaycarlton, line 1: I just now bumped this one fwiw.

jaycarlton (approved):

yonghaoy (approved):

google-storage/gradle.properties
wchamber: Thanks, I originally had the same, but pulled and updated so this is also getting bumped.

PR #88, opened 2021-05-18
PF-778 Migrate to Resource Manager v3

wchamber:
Migrate to the v3 Google Resource Manager API. This is a breaking change to the google-resourcemanager package as we migrate to the newest Google provided API and client library.

This migration is necessary to use both project & folder operations with the provided api client libraries. v1 & v2 are incomplete, see https://partnerissuetracker.corp.google.com/issues/137017056

This PR also adds some automatic prefixing of `projects/` to CloudResourceManagerCow.Projects methods. The new v3 API requires that the project id/number is always prefixed with `projects/`, and this PR allows this to be done silently as needed by CRL. This is continuing on the path of not only wrapping cloud APIs, but also adding some convenience.
We may want to consider doing this in other libraries where a simple prefix to a single thing is required, e.g. `serviceAccounts/`.


wchamber:
Github actions are down at the moment. I will make sure tests pass before merging.

google-cloudresourcemanager/src/main/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCow.java
```diff
@@ -186,17 +202,30 @@ private SetIamPolicy(CloudResourceManager.Projects.SetIamPolicy setIamPolicy) {
...
     }
   }
 
+  /**
+   * Helper function to prefix Project operation arguments with the required "projects/" prefix.
+   *
+   * <p>This allow CRL clients to directly pass project ids or numbers, or the already prefixed
+   * name.
+   */
+  private static String prefixProjects(String name) {
```
yonghaoy, line 222: Would it be helpful if we make this public to client?

yonghaoy (approved):
nice!

zloery (approved):

google-cloudresourcemanager/src/main/java/bio/terra/cloudres/google/cloudresourcemanager/CloudResourceManagerCow.java
wchamber: good idea, sure doesn't hurt.

PR #89, opened 2021-05-19
PF-774 include google-cloud-billing version and refresh gradle.lockfiles

wchamber:
We accidentally removed the google-cloud-billing version in #85. This causes compile errors downstream.
Also regenerate some of the gradle.lockfiles. It's mysterious how they got in this state or why they compile without issue, but we probably want them back.

This allows WSM to compile with local artifacts for me in https://github.com/DataBiosphere/terra-workspace-manager/pull/350


jaycarlton (approved):
Is there any way to add a test to catch something like this before it impacts downstream services? Can we set gradle to be more strict about dependency versions and disallow dep entries without a version specified?

wchamber:
> Is there any way to add a test to catch something like this before it impacts downstream services? Can we set gradle to be more strict about dependency versions and disallow dep entries without a version specified?

Sorry, I don't know.

ddietterich (approved):
LGTM

PR #95, opened 2021-06-01
Create CRL Gradle Platform to Align Library Versions

jmczerk:
The main intent of this PR is to create a [Gradle Platform](https://docs.gradle.org/current/userguide/java_platform_plugin.html) to allow for aligning versions of individual CRL libraries into a single versioned BOM for use downstream.

However, to get there I needed to do some Gradle refactoring to not depend on the use of the "discouraged" ([ref](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html)) `subprojects{}` construct, instead defining and using a Convention Plugin for the library sub-projects.  This is the bulk of the changes in the PR, and is broken out into its own commit.

To see how this would be consumed downstream: https://github.com/DataBiosphere/terra-workspace-manager/commit/7d60a4d1ef492562be0918a28ca3713dad0db391


buildSrc/src/main/groovy/terra-cloud-resource-lib.library-conventions.gradle
```diff
@@ -0,0 +1,118 @@
...
+    useJUnitPlatform {
+        includeTags "integration"
+    }
+    shouldRunAfter test
+}
+
+java {
+    // Builds Javadoc into the published package as part of the 'assemble' task.
+    withJavadocJar()
+    // Builds sources into the published package as part of the 'assemble' task.
```
jaycarlton, line 97: Nice. this wasn't working before AFAICT.

buildSrc/src/main/groovy/terra-cloud-resource-lib.library-conventions.gradle
```diff
@@ -0,0 +1,118 @@
...
+    withSourcesJar()
+}
+
+task minniekenny(type: Exec) {
```
jaycarlton, line 101: Maybe comment what this is for. I can never remember.

buildSrc/src/main/groovy/terra-cloud-resource-lib.library-conventions.gradle
```diff
@@ -0,0 +1,118 @@
+plugins {
```
jaycarlton, line 1: maybe comment up here about what a conventions plugin is and how this is a private plugin just for this project, etc.

jaycarlton (approved):
Thanks for doing this! I think this is a big step forward.

platform/gradle.properties
```diff
@@ -0,0 +1 @@
+version = 0.1.0
```
wchamber, line 1: Maybe let's also add a short README for the platform directory about what it's for and how it should be included/depended on by clients' gradle?

buildSrc/build.gradle
```diff
@@ -0,0 +1,21 @@
+// Gradle's recommended way of sharing build logic between subprojects is to use
+// "convention plugins" to capture shared build logic among different classes of
+// components.  This is prefered over the "discouraged" use of the
+// "subprojects{}" and "allprojects{}" constructs.  For more info please see:
+//
+// https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html
```
wchamber, line 6: maybe also link https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources ?

buildSrc/src/main/groovy/terra-cloud-resource-lib.library-conventions.gradle
wchamber: I don't get the distinction between a `library-conventions` and a `java-conventions` from here https://docs.gradle.org/current/samples/sample_convention_plugins.html. Do we only have a `library-conventions` here for a particular reason?

buildSrc/src/main/groovy/terra-cloud-resource-lib.library-conventions.gradle
```diff
@@ -0,0 +1,118 @@
...
+    api group: 'com.google.auth', name: 'google-auth-library-oauth2-http'
+    api group: 'com.google.cloud', name: 'google-cloud-core-http'
+
+    implementation group: 'org.slf4j', name: 'slf4j-api', version: '1.7.28'
+    annotationProcessor group: 'com.google.auto.value', name: 'auto-value', version: '1.7.3'
+    implementation group: 'com.google.auto.value', name: 'auto-value-annotations', version: '1.7.3'
+    implementation group: 'com.google.code.gson', name: 'gson', version: '2.8.6'
+    api group: 'com.google.guava', name: 'guava'
+
+    implementation group: 'bio.terra', name: 'terra-resource-janitor-client', version: janitorclient
```
wchamber, line 46: Should this be a dependency or a constraint? I don't really grok it.https://docs.gradle.org/current/samples/sample_convention_plugins.html

buildSrc/src/main/groovy/terra-cloud-resource-lib.library-conventions.gradle
```diff
@@ -0,0 +1,118 @@
...
+    ext {
+        janitorclient = '0.4.0-SNAPSHOT'
+    }
+    // Google dependencies - make sure we are using -jre and not android and use common bom
+    constraints {
+        implementation 'com.google.guava:guava:30.1.1-jre' // "-jre" for Java 8 or higher
+        testFixturesImplementation 'com.google.guava:guava:30.1.1-jre' // "-jre" for Java 8 or higher
+    }
+    implementation platform('com.google.cloud:libraries-bom:20.2.0')
+    testFixturesImplementation platform('com.google.cloud:libraries-bom:20.2.0')
```
wchamber, line 35: if we're using these, should we still be setting com.google.cloud library versions in the subprojects, e.g. google-storage?
Is there a similar bom platform for com.google.core or com.google.api-services?

(Ditto I still don't understand dependencies vs constraints for this.)

wchamber (commented):
thanks! few questions about what's going on.

buildSrc/src/main/groovy/terra-cloud-resource-lib.library-conventions.gradle
jmczerk: So my read of that doc (or one similar to it that I saw) is that if one had a project with both libraries and applications (and presumably several of each), they would have a `java-conventions` that applied to both libraries and applications, then a `library-conventions` for the shared library-building-specific stuff, and an `application-conventions` for the shared application-building-stuff; the `library-conventions` and `application-conventions` would both contain the `java-conventions` plugin.

Since all of the Java code in this repo is for a library, we could really call it `java-conventions` or `library-conventions`; I didn't feel strongly about either, so am open to changing.

buildSrc/src/main/groovy/terra-cloud-resource-lib.library-conventions.gradle
jmczerk: There's more detail here, but the main idea is that "constraints" impose constraints on which version of a component will be used if and only if that component is declared as a dependency (directly of transitively; thus this guava constraint was needed to make sure no transitive dependencies pulled the androind version).

Dependencies here are explicit dependencies; versions are omitted if a constraint is already applied, either explicitly or within a platform (and declaring a version that conflicts with an applied constraint will cause an error).

I will add something to the README.md to try to clarify this; the other thing of interest when declaring our platform is that the constraints do not imply dependencies, i.e by depending on the platform, one is not pulling all the CRL libraries as dependencies; it's just constraining which versions to use if one does (explicitly) take them as dependencies.

buildSrc/src/main/groovy/terra-cloud-resource-lib.library-conventions.gradle
jmczerk: If you need the library, it's a dependency.  A constraint would simply say "I only support version X of component Y"; it's really more useful for enforcing transitive dependencies in this context (i.e. if you were not using janitor, but wanted to make sure any transitive dependencies that did depend on it used a specific version).

buildSrc/src/main/groovy/terra-cloud-resource-lib.library-conventions.gradle
jmczerk: Hmmm, I just copy-pasted from the `subproject{}` block, so I'm not sure I helped anything here.  Is there something I should run to make sure it's working as expected?

jaycarlton: In IJ in a consuming project, see if you get actual sources with javadoc for the CRL  libs or just decompiled source. It seems like it either didn't pick up the sources jar by default or it wasn't there IIRC.

buildSrc/src/main/groovy/terra-cloud-resource-lib.library-conventions.gradle
jmczerk: @wchamber please let me know if my comment and/or the documentation I added helps clarify; if there are still questions we can discuss out-of-band.

buildSrc/src/main/groovy/terra-cloud-resource-lib.library-conventions.gradle
```diff
@@ -0,0 +1,133 @@
...
+
+configurations {
+    testImplementation.extendsFrom testFixturesImplementation
+}
+
+dependencies {
+    ext {
+        janitorclient = '0.4.0-SNAPSHOT'
+    }
+    // Google dependencies - make sure we are using -jre and not android and use common bom
```
jmczerk, line 42: One other thing to consider is that we could include this constraint in our platform (which would also pull a dependency on `com.google.cloud:libraries-bom:20.2.0`), so downstream would not have to do lines 43-48 anymore.

Docs on sourcing one platform from another platform [here](https://docs.gradle.org/current/userguide/java_platform_plugin.html#sec:java_platform_bom_import).

CC: @ddietterich

common/gradle.properties
```diff
@@ -1 +1,5 @@
+# When updating verion, please also consider updating the corresponding version
+# constraint with the new version in /platform/build.gradle and updating the
+# platform version in /platform/gradle.properties accordingly, following
+# semantic versioning.
```
wchamber, line 4: Maybe we should just link to our instructions on versioning?

platform/README.md
```diff
@@ -0,0 +1,31 @@
+# Cloud Resource Library Gradle Platform
```
wchamber, line 1: This is all great documentation. I think we might want to move it to the top level README, or some other top level .md if we start splitting things out. I wouldn't otherwise think to look in `platform` for this info.

wchamber (approved):
Thanks, this looks really nice.

buildSrc/src/main/groovy/terra-cloud-resource-lib.library-conventions.gradle
wchamber: Having CRL control the com.google.cloud bom makes sense to me - it's driving the cloud library support. I'm less sure it should do guava.

jmczerk: I believe that to safely declare _dependency_ on `com.google.cloud:libraries-bom`, we must provide the _constraints_ on `guava` to protect ourselves from bad transitive dependencies.  So I think we either need to do both or neither; I think @ddietterich understands the implications best.

platform/README.md
jmczerk: Good call...  I moved it into the main README, updated the publishing instructions to include the platform, and linked to publishing instructions in sub-project gradle.properties per your suggestion (I think the reminder is useful).

buildSrc/src/main/groovy/terra-cloud-resource-lib.library-conventions.gradle
ddietterich: Right. The `libraries-bom` is not specific about android vs jre versions of guava. We need to nail ourselves down to the jre version.

One surprise to me: the `libraries-bom` does not specify versions of very many libraries. If I understand correctly, it is specifying coherent versions of the common dependencies across libraries. So it doesn't include `storage`, but it includes `guava`.

PR #98, opened 2021-08-23
[IA-2860] Add Azure resource deployments

rtitle:
See: https://broadworkbench.atlassian.net/browse/IA-2860

This PR adds 3 CRL modules for Azure resources. The CRL modules mirror the modules in the Azure ResourceManager Java SDK: https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/resourcemanager

* `azure-resourcemanager-resources` -- manages deployments from ARM templates
* `azure-resourcemanager-compute` -- manages compute resources
* `azure-resourcemanager-common` -- common functionality for Azure modules

Also includes 1 (disabled) integration tests which:
* Authenticates as a publisher identity
* Deploys a Data Science VM template in the consumer's managed resource group
* Verifies the deployment succeeds
* Deletes the deployed resources

It works!

Still TODO:
* String constants are stubbed out (ping Rob for some real values). How should we handle Azure integration test dependencies/seed data?
* Should we integrate Azure resources with Janitor?
* Resource deletion could use some work. Right now the test nukes all resources in the managed resource group but there's probably a need to delete things more granularly based on tags.
* Test with new templates as we update them in https://github.com/DataBiosphere/terra-azure-arm-templates


azure-resource-manager/src/main/java/bio/terra/cloudres/azure/resouce/manager/AzureResourceConfiguration.java
```diff
@@ -0,0 +1,47 @@
+package bio.terra.cloudres.azure.resouce.manager;
```
mcovarr, line 1: resouce => resource

azure-resource-manager/src/test/resources/azuredeploy.json
```diff
@@ -0,0 +1,308 @@
+{
```
Qi77Qi, line 1: can the test reference a template from https://github.com/DataBiosphere/terra-azure-arm-templates/tree/main/leonardo instead of committing this one in this repo?

settings.gradle
```diff
@@ -2,6 +2,6 @@ rootProject.name = 'terra-cloud-resource-lib'
 include 'cloud-resource-schema', 'common', 'google-api-services-common', 'google-bigquery',
         'google-billing', 'google-cloudresourcemanager', 'google-compute', 'google-dns',
         'google-iam', 'google-notebooks', 'google-serviceusage','google-storage',
-        'platform'
+        'platform', 'azure-resource-manager'
```
Qi77Qi, line 5: it's essentially an ARM template deployment helper right?

azure-resource-manager/src/test/resources/azuredeploy.json
```diff
@@ -0,0 +1,308 @@
...
+      "type": "string",
+      "metadata": {
+        "description": "Username for Administrator Account"
+      }
+    },
+    "vmName": {
+      "type": "string",
+      "defaultValue": "vmName",
+      "metadata": {
+        "description": "The name of you Virtual Machine."
```
mcovarr, line 15: you => your

azure-resource-manager/src/test/resources/azuredeploy.json
rtitle: Yeah good idea, that probably makes more sense 👍

settings.gradle
rtitle: ARM == Azure Resource Manager :) 

`azure-resource-manager` is basically the Azure resource SDK. It can deploy templates (JSON) or individual resources via Java SDK. Right now in this PR there are only template deployments, but I think we'll end up having direct Java SDK calls too. I feel like it can all live under the `azure-resource-manager` module (unless we find a good reason to split them..)

rtitle: https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/overview

Qi77Qi: > I feel like it can all live under the azure-resource-manager module (unless we find a good reason to split things into separate modules..).

Yea, I wasn't sure if this module is intended to be an "all-in-one" module, or an `ARM` only module similar to the other modules in this repo.

this repo seems to split things, so it might be confusing if we bundle everything in this module...altho maybe we don't really need much more than ARM since ARM template is pretty powerful

azure-resource-manager/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+    api project(':common')
+
+    test {
+        useJUnitPlatform()
+    }
+
+    // https://mvnrepository.com/artifact/com.azure/azure-identity
+    api group: 'com.azure', name: 'azure-identity', version: '1.3.5'
+    // https://mvnrepository.com/artifact/com.azure.resourcemanager/azure-resourcemanager-resources
+    implementation group: 'com.azure.resourcemanager', name: 'azure-resourcemanager', version: '2.7.0'
```
Qi77Qi, line 17: is https://github.com/Azure/azure-sdk-for-java the github repo? might be useful to link the repo as well

settings.gradle
rtitle: Yeah.. maybe it makes sense to split the CRL modules based on Azure library, similar to GCP.

E.g. the template deployment stuff is (confusingly) in `azure-resourcemanager-resources` (https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/resourcemanager/azure-resourcemanager-resources).

Whereas compute calls are in `azure-resourcemanager-compute` (https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/resourcemanager/azure-resourcemanager-compute).

Will try to organize this…

rtitle:
Thanks @yonghaoy! I'll do another pass on this PR to try and clean some of this stuff up, then maybe we can discuss more from there.

azure-resource-manager/src/main/java/bio/terra/cloudres/azure/resouce/manager/AzureResourceConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+import com.azure.core.management.AzureEnvironment;
+import com.azure.core.management.profile.AzureProfile;
+import com.azure.resourcemanager.AzureResourceManager;
+
+import java.util.UUID;
+
+/**
+ * Configuration for working with Azure Resource Manager (ARM).
+ */
+public class AzureResourceConfiguration {
```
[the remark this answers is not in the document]
jdcanas: FYI, I wrote some of this code.

This is mainly here because muscles (@nmalfroy) had some code I used as a boilerplate. I'll be using this code shortly in a subsequent PR. 

That being said, I included it because I thought it was a good wrapper. There are different ways too configure an `AzureResourceManager` with various tenants (i.e. the home or user tentants) and subscriptions, so it felt good to have a class with readable methods. This could theoretically be handled a bit clearer with various constructors, but this class follows a common pattern we use where we have an object/class whose purpose is to create a client and handle initial config. See [this](https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocService.scala#L99) in workbench-libs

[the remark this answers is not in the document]
jdcanas: Hmm.. though I see @rtitle isn't using it as such here https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/98/files#diff-c4c51513c4cd784e9d8bc0cd8b7bc114fdb5e4516449db964cb6842894b5e339R51 and is creating the client directly instead of using `azureResourceConfiguration.getClient()`. We should probably make the constructor private/protected, and maybe make `getClient` more descriptive.

rtitle:
@wchamber I think this is ready for some eyes, interested to hear your thoughts.

azure-resourcemanager-resources/src/test/java/bio/terra/cloudres/azure/resourcemanager/resources/ResourceManagerCowTest.java
```diff
@@ -0,0 +1,156 @@
...
+  @AfterEach
+  public void afterEach() throws IOException {
+    cleanResourceGroup();
+  }
+
+  @Test
+  public void deployDsvmTemplate() throws IOException {
+    final URL templateUri =
+        new URL(
+            "https://raw.githubusercontent.com/DataBiosphere/terra-azure-arm-templates/main/leonardo/datascience-vm-customization/azuredeploy.json");
```
wchamber, line 67: I don't think CRL should be testing a "real" deployment template. A CRL test should have a dummy deployment template for the purposes of testing so that we're not broken by changes to an external repo.
Whether the template referenced here is a good one is a test for that repo, not CRL. CRL wants to be agnostic about its use.

azure-resourcemanager-resources/src/test/java/bio/terra/cloudres/azure/resourcemanager/resources/ResourceManagerCowTest.java
```diff
@@ -0,0 +1,156 @@
...
+            "Microsoft.Network/networkSecurityGroups"),
+        resourceTypesDeployed);
+    assertEquals(
+        Set.of("Succeeded"),
+        finalResult.deploymentOperations().list().stream()
+            .map(DeploymentOperation::provisioningState)
+            .collect(Collectors.toSet()));
+  }
+
+  private void cleanResourceGroup() throws IOException {
```
wchamber, line 123: Not this PR: This will inevitably leak resources during development, but fine for initial PR. Let's talk about using Janitor for cleanup.

wchamber (changes requested):

rtitle:
Thanks for the review @wchamber! I'll process this/respond tomorrow or Fri.

azure-resourcemanager-resources/src/main/java/bio/terra/cloudres/azure/resourcemanager/resources/ResourceManagerCow.java
```diff
@@ -0,0 +1,152 @@
...
+                .define(deploymentName)
+                .withExistingResourceGroup(resourceGroup)
+                .withTemplate(template)
+                .withParameters(
+                    parameters.entrySet().stream()
+                        .collect(
+                            Collectors.toMap(
+                                Map.Entry::getKey, e -> Map.of("value", e.getValue()))))
+                .withMode(deploymentMode)
+                .create(),
```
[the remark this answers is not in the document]
rtitle: This is really helpful, thank you! Will look into options.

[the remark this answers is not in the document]
rtitle: There is one thing I struggled with trying to wrap the Azure fluent API. For Google (`com.google.api.services`), all the operations (such as `Bigquery.Datasets.Get`) extend from `AbstractGoogleClientRequest<T>`. So we can decorate everything with `AbstractRequestCow<T>` which _wraps_ `AbstractGoogleClientRequest<T>` and plug in our logging.

For the Azure fluent API, I couldn't find such a common base class for _all_ operations. This makes it harder to just decorate things and inject `OperationAnnotator::exececuteCloudOperation` for cloud calls. There might be a creative way to do this though, need to spend more time with it.

I totally take the point of not hiding the cloud API, and just decorating it with our logging. Will see if I can make this work for Azure, also thinking a "CRL Azure Support" design doc would be useful since CRL is a bit more subtle than I thought initially. :)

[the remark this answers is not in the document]
rtitle: I guess there are probably a small set of base classes (`Creatable`, `Deletable`) that we could wrap with Cows…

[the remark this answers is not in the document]
wchamber: Yep, a doc SG.
\For GCP, it's not true that "everything' has a common base class, just the very raw auto generated libraries from `com.google.api.services`. google-storage, for example, has a `com.google.cloud` hand crafted library. There, we're wrapping objects that have cloud call methods on them without any base class, e.g. `BlobCow` and `BucketCow`. But that let's extreme than the number of classes to wrap for the fluent API. https://github.com/DataBiosphere/terra-cloud-resource-lib/tree/master/google-storage/src/main/java/bio/terra/cloudres/google/storage

rtitle:
Per offline discussion, I think I'm going to close this PR. Will follow up with a design doc w.r.t. CRL Azure support.

PR #99, opened 2021-09-14
[IA-2860] Azure models/tests for Network / Disk / VM resources CREATE / GET / DELETE

jdcanas:
This is organized into the addition of 3 new models with a base model. It adds a test for disks and network in isolation, and a test to create all resources (IP, network, Disk), and then attach these resources to a vm. It's possible to create a VM with new resources too, but it seemed more logical to leverage the fact that the code for creating the components individually was already tested. 

Its worth noting this test creates a VM synchronously, which causes the tests to take ~4minutes. It's possible that we will want to handle this async in the application, which the azure SDK supports natively along with a built-in polling interface, but it made more sense to do it sync for the tests since verification is required and blocking before subsequent steps.

Note that this is not merging to or based off main, it depends on the code in `rt-azure`, which provides the base models and Context/Cow setup, among other things.


rtitle (approved):
Looks good to merge into my branch - thanks for the assist! I'll clean up, etc then make the full PR into `main`. 

Looks like there are some conflicts to be addressed before merging though.

PR #100, opened 2021-09-14
[IA-2860] Add CRL code to create/delete Azure VMs

rtitle:
JIRA: https://broadworkbench.atlassian.net/browse/IA-2860

See design doc for a description of the overall approach: https://docs.google.com/document/d/170AaSyYdifbrgNZ543_v0NgNCEhX00F9NryFpBWIYQc/edit

Specifically this PR adds:
* `azure-resourcemanager-common` and `azure-resourcemanager-compute` modules
* `ComputeManagerCow` class which wraps Azure `ComputeManager`, providing standard logging options
* `AzureRequestLogger` and `AzureResponseLogger` classes to integrate with Janitor and Terra structured logging, respectively
* An integration test which creates a public IP, disk, network, subnet, and a VM. It passes!

There are a couple things left to do, but I think can be done outside of this PR:
1. Add test environment data to vault and update `render-test-configs.sh`. So far I've been manually providing my own environment. The integration test is disabled because of this.
2. Add Janitor-side cleanup code. Currently there are hooks for calling Janitor but it's stubbed out with a TODO. It looks like Janitor needs to consume the new `cloud_resources_uid.yaml` schema after this PR merges.

My goal is to merge this to the CRL `master` branch, and then use from the WSM `azure-poc` branch. We can iterate by adding more Azure APIs as needed. But please let me know of any feedback.


azure-resourcemanager-common/src/main/java/bio/terra/cloudres/azure/resourcemanager/common/AzureRequestLogger.java
```diff
@@ -0,0 +1,42 @@
...
+/** Intercepts Azure HTTP requests to record cloud resource creations for cleanup. */
+public class AzureRequestLogger implements HttpRequestLogger {
+  private final ClientConfig clientConfig;
+
+  AzureRequestLogger(ClientConfig clientConfig) {
+    this.clientConfig = clientConfig;
+  }
+
+  @Override
+  public Mono<Void> logRequest(ClientLogger logger, HttpRequestLoggingContext loggingOptions) {
```
rtitle, line 23: This is the place where we call Janitor for resource creation requests.

azure-resourcemanager-common/src/main/java/bio/terra/cloudres/azure/resourcemanager/common/AzureResponseLogger.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+public class AzureResponseLogger implements HttpResponseLogger {
+  private final ClientConfig clientConfig;
+
+  AzureResponseLogger(ClientConfig clientConfig) {
+    this.clientConfig = clientConfig;
+  }
+
+  @Override
+  public Mono<HttpResponse> logResponse(
```
rtitle, line 32: This is where we log structured data for API calls.

azure-resourcemanager-compute/src/main/java/bio/terra/cloudres/azure/resourcemanager/compute/data/CreateDiskRequestData.java
```diff
@@ -0,0 +1,30 @@
...
+  private final int size;
+
+  public CreateDiskRequestData(String resourceGroupName, String name, Region region, int size) {
+    super(ComputeManagerOperation.AZURE_CREATE_DISK, resourceGroupName, name, region);
+    this.size = size;
+  }
+
+  @Override
+  public Optional<CloudResourceUid> resourceUidCreation() {
+    // TODO: populate this when the CloudResourceUid Janitor model is regenerated
```
rtitle, line 20: We can update this when Janitor consumes the new `cloud_resources_uid.yaml` schema and generates the Java objects.

(It's a bit of a circular dependency: CRL needs to update the schema, to regenerate the classes in Janitor, which are needed in CRL.)

render-test-config.sh
```diff
@@ -22,4 +22,15 @@ docker run --rm -e VAULT_TOKEN=$VAULT_TOKEN ${DSDE_TOOLBOX_DOCKER_IMAGE} \
 docker run --rm --cap-add IPC_LOCK \
             -e VAULT_TOKEN=$VAULT_TOKEN ${DSDE_TOOLBOX_DOCKER_IMAGE} \
             vault read -format json ${VAULT_SERVICE_ACCOUNT_JANITOR_CLIENT_PATH} \
-            | jq -r .data.key | base64 -d > ${SERVICE_ACCOUNT_JANITOR_CLIENT_OUTPUT_FILE_PATH}
\ No newline at end of file
+            | jq -r .data.key | base64 -d > ${SERVICE_ACCOUNT_JANITOR_CLIENT_OUTPUT_FILE_PATH}
+
+# TODO(RT): specify Azure secrets in vault
```
rtitle, line 27: This is where we will read the Azure env data from vault.

azure-resourcemanager-common/src/testFixtures/java/bio/terra/cloudres/azure/resourcemanager/common/AzureIntegrationUtils.java
```diff
@@ -0,0 +1,112 @@
...
+import com.azure.core.credential.TokenCredential;
+import com.azure.core.management.AzureEnvironment;
+import com.azure.core.management.profile.AzureProfile;
+import com.azure.identity.ClientSecretCredentialBuilder;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+/** Utilities for Azure integration tests. */
+public class AzureIntegrationUtils {
```
ddietterich, line 12: This is a different approach than we have used in the past. Maybe checkout how [render-test-configs.sh](https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/render-test-config.sh#L1) works. We use a similar approach in WSM:
- the render step does the vault access and populates a set of files
- a Spring-profile based application properties file points to the file paths and provides other non-secure parameters.

azure-resourcemanager-common/src/testFixtures/resources/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
```
ddietterich, line 1: I know this is the style, but it is unnecessary to merge this empty directory into github. The directory itself can be ignored in the top-level .gitignore (all the more easily if it had a more distinct name!).

cloud-resource-schema/src/main/resources/cloud_resources_uid.yaml
```diff
@@ -72,6 +72,66 @@ components:
       properties:
         projectId:
           type: string
+    AzurePublicIp:
+      type: object
+      required:
```
ddietterich, line 77: I'm curious why these UIDs do not need the tenant and subscription to provide the "full path" to the object.

ddietterich (commented):
I gave this a look and added a few comments.

azure-resourcemanager-compute/src/main/java/bio/terra/cloudres/azure/resourcemanager/compute/data/CreateDiskRequestData.java
wchamber: Yes. We should remove the cloud_resources_uid from CRL and just make it a part of the janitor client for CRL to fill out. But we haven't done that yet.

azure-resourcemanager-common/src/main/java/bio/terra/cloudres/azure/resourcemanager/common/AzureRequestLogger.java
```diff
@@ -0,0 +1,42 @@
...
+import bio.terra.cloudres.common.cleanup.CleanupRecorder;
+import com.azure.core.http.policy.HttpRequestLogger;
+import com.azure.core.http.policy.HttpRequestLoggingContext;
+import com.azure.core.util.Context;
+import com.azure.core.util.logging.ClientLogger;
+import java.util.Optional;
+import reactor.core.publisher.Mono;
+
+/** Intercepts Azure HTTP requests to record cloud resource creations for cleanup. */
+public class AzureRequestLogger implements HttpRequestLogger {
```
wchamber, line 15: class name and comment don't seem to align for me. is this for logging all requests, or just resource creation requests?

azure-resourcemanager-common/src/main/java/bio/terra/cloudres/azure/resourcemanager/common/Defaults.java
```diff
@@ -0,0 +1,26 @@
...
+import bio.terra.cloudres.common.ClientConfig;
+import com.azure.core.http.policy.HttpLogDetailLevel;
+import com.azure.core.http.policy.HttpLogOptions;
+import com.azure.core.util.Context;
+
+/** Provides defaults for working with the Azure Resource Manager API. */
+public class Defaults {
+  private Defaults() {}
+
+  static final String CLOUD_RESOURCE_REQUEST_DATA_KEY = "requestData";
```
wchamber, line 12: nit: this is pretty abstract, maybe `"crlRequestData"`?

azure-resourcemanager-common/src/testFixtures/java/bio/terra/cloudres/azure/resourcemanager/common/AzureIntegrationUtils.java
wchamber: CRL has so far not included Spring, so we'd be doing it just for grabbing application property files in tests. Not a hard objection, just an observation.

azure-resourcemanager-compute/src/main/java/bio/terra/cloudres/azure/resourcemanager/compute/ComputeManagerCow.java
```diff
@@ -0,0 +1,40 @@
+package bio.terra.cloudres.azure.resourcemanager.compute;
+
+import bio.terra.cloudres.azure.resourcemanager.common.Defaults;
+import bio.terra.cloudres.common.ClientConfig;
+import com.azure.core.credential.TokenCredential;
+import com.azure.core.management.profile.AzureProfile;
+import com.azure.resourcemanager.compute.ComputeManager;
+
+/** A Cloud Object Wrapper (COW) for Azure {@link ComputeManager} library. */
+public class ComputeManagerCow {
```
wchamber, line 10: This doesn't seem to be wrapping much at all. AFAICT, we only care about making sure `ComputeManager.configure()` happens with some CRL specific options. So let's make a static function to do just that and leave everything else outside of CRL.
`static ComputeManager crlConfigure(ComputeManager manager, ClientConfig clientConfig)`
Maybe we only need one method that can handle every azure manager.

azure-resourcemanager-compute/src/test/java/bio/terra/cloudres/azure/resourcemanager/compute/ComputeManagerCowTest.java
```diff
@@ -0,0 +1,377 @@
...
+        .networkManager()
+        .publicIpAddresses()
+        .define(name)
+        .withRegion(defaultRegion)
+        .withExistingResourceGroup(resourceGroupName)
+        .withDynamicIP()
+        .withTag("crl.integration", "true")
+        .create(
+            Defaults.buildContext(
+                new CreatePublicIpRequestData(resourceGroupName, name, defaultRegion)));
```
wchamber, line 270: If I'm understanding this correctly, clients need to remember to always do this for every method. Seems ok, but if there were a way to register something in the logger that is able to pull out the interesting information for each api endpoint, that would save our clients some boilerplate. Is that feasible?

As-is, I don't really see the point of these integration tests. We're barely doing any wrapping, so none of our methods are really being executed in different ways for different endpoints. Having one endpoint that verifies our logger is hooked in correctly seems sufficient.

azure-resourcemanager-compute/src/main/java/bio/terra/cloudres/azure/resourcemanager/compute/data/BaseRequestData.java
```diff
@@ -0,0 +1,33 @@
...
+import bio.terra.cloudres.azure.resourcemanager.common.AbstractRequestData;
+import bio.terra.cloudres.common.CloudOperation;
+import com.azure.core.management.Region;
+import com.google.gson.JsonObject;
+
+/**
+ * Extends {@link AbstractRequestData} to add common fields for working with the Compute Manager
+ * API.
+ */
+public abstract class BaseRequestData extends AbstractRequestData {
```
wchamber, line 12: nit: maybe `ComputeBaseRequestData`? even with package scoping, `BaseRequestData` sounds pretty broad.

azure-resourcemanager-compute/src/main/java/bio/terra/cloudres/azure/resourcemanager/compute/data/BaseRequestData.java
```diff
@@ -0,0 +1,33 @@
...
+  protected final String name;
+  protected final Region region;
+  protected final String resourceGroupName;
+
+  protected BaseRequestData(
+      CloudOperation cloudOperation, String resourceGroupName, String name, Region region) {
```
wchamber, line 18: these constructors are going to get very long and very easy to mess up by taking many String primitives. I think we should start by writing builders for these data classes to make it harder to mix up a `resourceGroupName` and a `name`.
Consider using AutoValue to get easy builders https://github.com/google/auto/blob/master/value/userguide/builders.md

cloud-resource-schema/src/main/resources/cloud_resources_uid.yaml
rtitle: That is a good point, I think they do need tenant and subscription. I was thrown by the APIs all defining `.getByResourceGroup(resourceGroup, name)` but there is a tenant/subscription context. Will update.

azure-resourcemanager-common/src/main/java/bio/terra/cloudres/azure/resourcemanager/common/AzureResponseLogger.java
```diff
@@ -0,0 +1,67 @@
...
+            data -> {
+              AbstractRequestData requestData = (AbstractRequestData) data;
+              logData.addProperty("operation", requestData.cloudOperation().name());
+              logData.add("requestData", requestData.serialize());
+            });
+
+    logger.info(
+        "CRL completed Azure request",
+        // Include logData for terra-common-lib logging to pick up and include in JSON output.
+        logData);
```
wchamber, line 63: Baseline, I think it's valuable to use consistent logging/tracing/metrics between Azure and GCP. We need to rework the `OperationAnnotator` API and it might not be feasible to have apples to apples monitoring, but if we don't try at all we'll have subtly different regimes between azure and gcp.
Consider if there's a "log data" value class to use in CRL's `common` package and populate that here.
Consider if we can use that to both write logs and metrics in a consistent fashion.
Is there some way we can do tracing on Azure as well?

azure-resourcemanager-common/src/testFixtures/resources/.gitignore
rtitle: +1, makes sense.

azure-resourcemanager-compute/src/test/java/bio/terra/cloudres/azure/resourcemanager/compute/ComputeManagerCowTest.java
```diff
@@ -0,0 +1,377 @@
...
+            .fromAnyAddress()
+            .fromAnyPort()
+            .toAddress("INTERNET")
+            .toAnyPort()
+            .withAnyProtocol()
+            .attach()
+            .create(
+                Defaults.buildContext(
+                    new CreateNetworkSecurityGroupRequestData(
+                        resourceGroupName, subnetName, defaultRegion)));
```
wchamber, line 323: it seems like there are a lot of arguments for the request that are not being included in the request data to log. Do we want to leave that much out? Can we write down some guidelines on what's good to include/exclude in an arbitrary RequestData?

wchamber:
@yonghaoy  fyi, curious what you think.

azure-resourcemanager-common/src/testFixtures/java/bio/terra/cloudres/azure/resourcemanager/common/AzureIntegrationUtils.java
rtitle: Correct, I updated `render-test-configs.sh` to generate this `.properties` file. I was imagining `render-test-configs.sh` would read the data from vault eventually.

I feel like it would work with Spring profiles in WSM, by pointing to a different `.properties` file per Spring profile. But there is no Spring in CRL.

azure-resourcemanager-common/src/main/java/bio/terra/cloudres/azure/resourcemanager/common/AzureRequestLogger.java
rtitle: The hook fires for every http request, but it is a no-op unless the creation resource uid is present. I could make this more clear in comments.

azure-resourcemanager-compute/src/main/java/bio/terra/cloudres/azure/resourcemanager/compute/ComputeManagerCow.java
rtitle: Sure sounds good! The less wrapping the better. I can try that out.

azure-resourcemanager-compute/src/test/java/bio/terra/cloudres/azure/resourcemanager/compute/ComputeManagerCowTest.java
rtitle: 1. Yes you are correct. The logger does try to pull certain things out of every request, namely the http method (POST, GET, etc), request URL, status code, retry count, and duration. The `*RequestData` objects are really for optionally adding extra information to the logs depending on data type. 

Alternatively we could make the logger just log the entire request body. But I was a little nervous to do that -- I'm not sure it's predictable what format the body will be in, and whether it'll fit nicely within our structured logs (we may end up with a JSON string containing a giant escaped JSON string).

Idk, what do you think?

2. Re integration tests, yes I see your point. I was kind of using these tests as a way to develop the VM creation calls that we'll eventually use in WSM/Leo. So even if they are not technically exercising many different CRL code paths, they are still useful as examples of the Azure API (but I'm happy to move this somewhere else if you prefer).

azure-resourcemanager-common/src/main/java/bio/terra/cloudres/azure/resourcemanager/common/AzureResponseLogger.java
rtitle: Hm ok - will think about using common objects for GCP and Azure log data (or maybe a common interface with subclasses if there are slight differences?).

azure-resourcemanager-compute/src/test/java/bio/terra/cloudres/azure/resourcemanager/compute/ComputeManagerCowTest.java
rtitle: Yeah… might be good to have richer request data. Especially since most of what is request data now is already present in the request URL (which is also logged). Will do another pass on request data for each resource type.

rtitle:
Thanks @wchamber @yonghaoy and @ddietterich for the great reviews! I replied to some comments above, and I will push some changes to address the feedback. I'll re-ping when this is ready for another look.

azure-resourcemanager-compute/src/test/java/bio/terra/cloudres/azure/resourcemanager/compute/ComputeManagerCowTest.java
wchamber: 1. I think we at least want what's being pulled out in the logger already (http method, request url, status code etc.), but also maybe the entire request body. The entire request body might not be especially useful (even if we unescape the json properly) compared to the custom fields of the *RequestData objects. the full request body might be very verbose, but we could leave it as debug level log, or allow it's configuration based on request url patterns, etc.
I think custom RequestData objects are more work but higher yield. The entire json body is probably fine, but possibly too much data to dump on every request.

2. I don't think these should live in CRL, at least as integration tests. I'd like to reserve tests for exercising CRL code and not sign up for the maintenance burden of something else. I think it'd be fine to start an `examples` project in CRL or somewhere else if that's useful, but for my 2c I'd expect it to rot very quickly once we start having Azure API use in other repos.

azure-resourcemanager-common/src/main/java/bio/terra/cloudres/azure/resourcemanager/common/AzureRequestLogger.java
yonghaoy: I don't feel we need to implement `HttpRequestLogger` or we do not even need this class if just for Janitor. Feels the only reason we make it an request logger is because it can be auto fired, not for any functional reason. What if we just we just put `CleanupRecorder.record()` in corresponding creation method?

azure-resourcemanager-common/src/testFixtures/java/bio/terra/cloudres/azure/resourcemanager/common/AzureIntegrationUtils.java
```diff
@@ -0,0 +1,112 @@
...
+   */
+  public static TokenCredential getAdminAzureCredentialsOrDie() {
+    try (InputStream in =
+        Thread.currentThread().getContextClassLoader().getResourceAsStream(AZURE_PROPERTIES_PATH)) {
+      Properties properties = new Properties();
+      properties.load(in);
+
+      final String clientId = properties.getProperty(AZURE_PROPERTY_PREFIX + ".admin.clientId");
+      if (clientId == null) {
+        throw new RuntimeException(
```
yonghaoy, line 34: optional: I don't see too many benefits from check null then throw RuntimeException than just throw NPE or whatever errors later. 
Similar questions for all other throwing RuntimeExceptions scenarios, most of them seems unnecessary. 
Or we can just use `Preconditions. checkNotNull(clientId, "message");`

azure-resourcemanager-compute/src/main/java/bio/terra/cloudres/azure/resourcemanager/compute/ComputeManagerOperation.java
```diff
@@ -0,0 +1,12 @@
+package bio.terra.cloudres.azure.resourcemanager.compute;
+
+import bio.terra.cloudres.common.CloudOperation;
+
+/** {@link CloudOperation} for using Azure ComputeManager API. */
+public enum ComputeManagerOperation implements CloudOperation {
+  AZURE_CREATE_PUBLIC_IP,
```
yonghaoy, line 7: nit sort alphabetically?

azure-resourcemanager-compute/src/main/java/bio/terra/cloudres/azure/resourcemanager/compute/data/CreateVirtualMachineRequestData.java
```diff
@@ -0,0 +1,49 @@
...
+  private final Disk disk;
+
+  public CreateVirtualMachineRequestData(
+      String resourceGroupName,
+      String name,
+      Region region,
+      Network network,
+      String subnetName,
+      PublicIpAddress ip,
+      Disk disk) {
```
yonghaoy, line 26: Optional nit: Consider to use Builder patten when faced with many constructor parameter. 
https://github.com/david-sauvage/effective-java-summary#builder
Google autovalue is heavily used in some Terra APPs and it is pretty convenient

azure-resourcemanager-compute/src/test/java/bio/terra/cloudres/azure/resourcemanager/compute/ComputeManagerCowTest.java
```diff
@@ -0,0 +1,377 @@
...
+                    resourceGroupName, networkName, defaultRegion, subnetName)));
+  }
+
+  private static VirtualMachine createVM(
+      String name, Network network, String subnetName, PublicIpAddress ip, Disk disk) {
+    // Note: these are public credentials, taken from:
+    // https://github.com/Azure-Samples/network-java-manage-virtual-network/blob/master/src/main/java/com/azure/resourcemanager/network/samples/ManageVirtualNetwork.java
+    final String userName = "tirekicker";
+    final String sshKey =
+        "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCfSPC2K7LZcFKEO+/t3dzmQYtrJFZNxOsbVgOVKietqHyvmYGHEC0J2wPdAqQ/63g/hhAEFRoyehM+rbeDri4txB3YFfnOK58jqdkyXzupWqXzOrlKY4Wz9SKjjN765+dqUITjKRIaAip1Ri137szRg71WnrmdP3SphTRlCx1Bk2nXqWPsclbRDCiZeF8QOTi4JqbmJyK5+0UqhqYRduun8ylAwKKQJ1NJt85sYIHn9f1Rfr6Tq2zS0wZ7DHbZL+zB5rSlAr8QyUdg/GQD+cmSs6LvPJKL78d6hMGk84ARtFo4A79ovwX/Fj01znDQkU6nJildfkaolH2rWFG/qttD azjava@javalib.com";
```
yonghaoy, line 350: Is this a real secret?

azure-resourcemanager-common/src/main/java/bio/terra/cloudres/azure/resourcemanager/common/AzureResponseLogger.java
yonghaoy: It would be great if we can generalize annotator to support both. 
Logging: What is the benefit we get from using Azure's requestLogger? I think CRL will also log every request, and what we need is just those basic stuff. Also CRL's annotator logEvent a simpler pattern than introducing few new classes then hook them.
For the code above, do we need to check things inside `loggingOptions`, `request` then convert them to string? Can we just deserialize `loggingOptions`, `request`, and this can be just
`logData.addProperty("loggingOptions", new Gson().toJsonTree(loggingOptions));`

RE Tracing: The use case for tracing in CRL is also simple, there is no complicated logic for each cloud call, they are atomic. Using openness like in Annotator is enough.
For advanced tracing which is used by the Server that use CRL,  we have `https://github.com/DataBiosphere/terra-common-lib/tree/develop/src/main/java/bio/terra/common/tracing` does that, and it works well with CRL's tracing. I am not sure how `azure-core-tracing-opentelemetry` works, but I am concerned that it may not work well with our tracing library.

yonghaoy (commented):
Thanks! This looks awesome!

azure-resourcemanager-common/src/testFixtures/java/bio/terra/cloudres/azure/resourcemanager/common/AzureIntegrationUtils.java
rtitle: `Preconditions` sounds good

azure-resourcemanager-compute/src/test/java/bio/terra/cloudres/azure/resourcemanager/compute/ComputeManagerCowTest.java
rtitle: Heh no it's a dummy key pulled from an Azure example. The `Compute` builder checks the key format (e.g. you can't just passed "dummy-key").

However per an earlier comment I think I'm going to remove some of these integration tests -- they are not all really needed to cover CRL functionality (e.g. these are more examples of Azure API usage than CRL tests).

rtitle:
@wchamber @yonghaoy @ddietterich sorry for the delay, but I think I applied all PR feedback. It's mostly in [this commit](https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/100/commits/415db79a57599e44104cd3c0d54c384b9785720a). Summary of changes:

1. Generalize `OperationAnnotator` to work for both GCP and Azure. To facilitate this I added a `recordOperation` method and a `OperationData` value class.
2. Changed the `*RequestData` classes to use the `@AutoValue` builder pattern, and added some more fields. The particular fields we capture may be an iterative process as we start to use these APIs more in production, but hopefully what exists is a good starting point.
3. Removed unnecessary integration tests; also added some unit tests to verify serialization.
4. Some other cleanup as well.

PTA(nother)L - am hoping this is close! :)

azure-resourcemanager-common/src/main/java/bio/terra/cloudres/azure/resourcemanager/common/AzureRequestLogger.java
```diff
@@ -0,0 +1,46 @@
...
+import com.azure.core.util.logging.ClientLogger;
+import java.util.Optional;
+import reactor.core.publisher.Mono;
+
+/**
+ * Intercepts Azure HTTP requests to record cloud resource creations for cleanup.
+ *
+ * <p>This is a no-op if the HTTP request is not a cloud resource creation.
+ */
+public class AzureRequestLogger implements HttpRequestLogger {
```
wchamber, line 19: Change the class name to reflect that this is for logging cloud resource creation.

azure-resourcemanager-common/src/main/java/bio/terra/cloudres/azure/resourcemanager/common/AzureResponseLogger.java
```diff
@@ -0,0 +1,134 @@
...
+    final long contentLength;
+    try {
+      contentLength = Long.parseLong(contentLengthString);
+    } catch (NumberFormatException | NullPointerException e) {
+      return;
+    }
+
+    // The body is logged if the Content-Type is not "application/octet-stream" and the body isn't
+    // empty
+    // and is less than 16KB in size.
```
wchamber, line 115: nit comment whitespacing

azure-resourcemanager-common/src/main/java/bio/terra/cloudres/azure/resourcemanager/common/ResourceManagerOperation.java
```diff
@@ -0,0 +1,12 @@
...
+
+import bio.terra.cloudres.common.CloudOperation;
+
+/**
+ * {@link CloudOperation} for using Azure ResourceManager API.
+ *
+ * <p>This is used as a generic fallback; in practice, a more specific value should be used.
+ */
+public enum ResourceManagerOperation implements CloudOperation {
+  AZURE_RESOURCE_MANAGER_OPERATION
```
wchamber, line 11: can we make the enum explicitly default or generic rather than having a vague name? brainstorming, `AZURE_RESOURCE_MANAGER_UNKNOWN_OPERATION` `_GENERIC_`

azure-resourcemanager-compute/src/main/java/bio/terra/cloudres/azure/resourcemanager/compute/data/CreateDiskRequestData.java
```diff
@@ -0,0 +1,43 @@
...
+
+import bio.terra.cloudres.azure.resourcemanager.compute.ComputeManagerOperation;
+import bio.terra.cloudres.common.CloudOperation;
+import com.azure.core.management.Region;
+import com.google.auto.value.AutoValue;
+import com.google.gson.JsonObject;
+
+/** Disk creation request data. */
+@AutoValue
+public abstract class CreateDiskRequestData extends BaseComputeRequestData {
```
wchamber, line 11: up to you; consider using composition instead of inheritance for `BaseComputeRequestData`. If every compute `RequestData` could have-a `CommonComputeRequestData` instead of being-a `BaseComputeRequestData`. The common data could have it's own builder rather than copying its fields into a subclass's builder each time.

azure-resourcemanager-compute/src/main/java/bio/terra/cloudres/azure/resourcemanager/compute/data/BaseComputeRequestData.java
```diff
@@ -0,0 +1,29 @@
...
+  public abstract String name();
+
+  /** The region of the resource. */
+  public abstract Region region();
+
+  /** The resource group of the resource. */
+  public abstract String resourceGroupName();
+
+  @Override
+  public JsonObject serialize() {
```
wchamber, line 22: this seems error-prone. If there's a subclass of BaseComputeRequestData, it should override serialize. With this, it's possible to forget that. Could use composition, or make a protected `serializeCommon` helper methods to be used by subclasses.

azure-resourcemanager-compute/src/test/java/bio/terra/cloudres/azure/resourcemanager/compute/ComputeManagerTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.jupiter.api.Test;
+
+/**
+ * Integration test for {@link ComputeManager} using CRL.
+ *
+ * <p>This test does not attempt to fully cover {@link ComputeManager} functionality, but creates
+ * some representative resources to exercise CRL functionality.
+ */
+@Tag("integration")
+// Note: temporarily disabled because we have not yet added Azure test environment setup to vault.
```
wchamber, line 31: Did we agree to remove these from CRL since they're not testing CRL code?

azure-resourcemanager-compute/src/test/java/bio/terra/cloudres/azure/resourcemanager/compute/ComputeRequestDataTest.java
```diff
@@ -0,0 +1,133 @@
...
+            .setRegion(Region.US_EAST)
+            .setAddressSpaceCidr("192.168.0.0/16")
+            .setSubnetName("my-subnet")
+            .setAddressPrefix("192.168.1.0/24")
+            .setNetworkSecurityGroup(mockNetworkSecurityGroup())
+            .build();
+
+    assertEquals(ComputeManagerOperation.AZURE_CREATE_NETWORK, createNetwork.cloudOperation());
+    assertEquals(
+        "{\"resourceGroupName\":\"my-rg\",\"name\":\"my-network\",\"region\":\"eastus\",\"addressSpaceCidr\":\"192.168.0.0/16\",\"subnetName\":\"my-subnet\",\"addressPrefix\":\"192.168.1.0/24\",\"networkSecurityGroupName\":\"my-nsg\"}",
```
wchamber, line 52: nit: add some linebreaks with string `+` https://google.github.io/styleguide/javaguide.html#s4.4-column-limit
ditto others.

cloud-resource-schema/src/main/resources/cloud_resources_uid.yaml
```diff
@@ -72,6 +72,106 @@ components:
...
+        - tenantId
+        - subscriptionId
+        - resourceGroupName
```
wchamber, line 80: Are these common to all Azure resources? If so, consider making a grouping object to re-use in all resources.

common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -64,39 +67,63 @@ public OperationAnnotator(ClientConfig clientConfig, Logger logger) {
...
       span.end();
     }
   }
 
+  /**
+   * Records log, tracing, and metric data associated with a cloud operation.
+   *
+   * @param operationData the {@link OperationData} to record.
+   */
+  public void recordOperation(OperationData operationData) {
```
wchamber, line 106: worth a note comparing when to use this vs the executeCowOperationMethods

common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -119,37 +146,37 @@ private void recordLatency(Duration duration, CloudOperation operation) {
    * <p>If an exception is present, it will also be included as a logging argument. SLF4J should
    * recognize this argument and include a stacktrace in the resulting error log message.
```
wchamber, line 147: nit: touch up this and the comment before it to refer to OperationData

common/src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
```diff
@@ -138,10 +139,13 @@ public void testLogEvent() throws Exception {
     operationAnnotator = new OperationAnnotator(clientConfig, mockLogger);
 
     operationAnnotator.logEvent(
-        StubCloudOperation.TEST_OPERATION,
-        PROJECT_REQUEST,
-        Duration.ofMillis(2345),
-        Optional.empty());
+        OperationData.builder()
```
wchamber, line 142: add tests for `recordEvent` as we have tests for the `executeCowOperation` methods

render-test-config.sh
wchamber: A tracking ticket would be nice.

wchamber (approved):
thanks!

azure-resourcemanager-compute/src/main/java/bio/terra/cloudres/azure/resourcemanager/compute/data/CreateDiskRequestData.java
rtitle: Yeah maybe composition would be better..

azure-resourcemanager-compute/src/test/java/bio/terra/cloudres/azure/resourcemanager/compute/ComputeManagerTest.java
rtitle: I removed most of the test cases but left 1. I think it's valid to exercise the `crlConfigure` method and verify it returns a usable Azure `ComputeManager` instance. Maybe I can reframe the test for that purpose (e.g. call it `testCrlConfigure` and then just make some representative calls). wdyt?

cloud-resource-schema/src/main/resources/cloud_resources_uid.yaml
rtitle: Thought about that but didn't because the existing file has a pretty flat structure. But should be okay to make a common model object..

common/src/test/java/bio/terra/cloudres/common/OperationAnnotatorTest.java
rtitle: Thanks, good catch

common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
```diff
@@ -64,39 +67,63 @@ public OperationAnnotator(ClientConfig clientConfig, Logger logger) {
       CloudOperation cloudOperation, CowCheckedExecute<R, E> cowExecute, CowSerialize cowSerialize)
```
yonghaoy, line 67: I guess it would be hard to let Azure use `executeCowOperation` directly, so we make Azure packages uses recordOperation instead? But if that is possible, I guess we would not need AzureReuqest/ResponseLogger.

yonghaoy (approved):

common/src/main/java/bio/terra/cloudres/common/OperationData.java
```diff
@@ -0,0 +1,50 @@
...
+    return new AutoValue_OperationData.Builder();
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
+    public abstract Builder setDuration(Duration duration);
+
+    public abstract Builder setTryCount(OptionalInt tryCount);
+
+    public abstract Builder setExecutionException(Optional<Exception> executionException);
```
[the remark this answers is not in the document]
rtitle: Oh cool, didn't realize the special `Optional` handling

azure-resourcemanager-compute/src/test/java/bio/terra/cloudres/azure/resourcemanager/compute/ComputeManagerTest.java
wchamber: Sounds ok. In that case, I would expect kind of the minimum number of calls, maybe a get or a create/delete. I don't know what checking that get & list both work does.
I'd kind of expect something in the test that would fail if we didn't use `crlConfigure` or `Defaults.buildContext`, but maybe a single test for that somewhere central instead of a test in every package (e.g. `compute`) would be better.

azure-resourcemanager-common/src/main/java/bio/terra/cloudres/azure/resourcemanager/common/Defaults.java
```diff
@@ -0,0 +1,44 @@
...
+import com.azure.resourcemanager.resources.fluentcore.arm.AzureConfigurable;
+
+/** Provides defaults for working with the Azure Resource Manager API. */
+public class Defaults {
+  private Defaults() {}
+
+  static final String CLOUD_RESOURCE_REQUEST_DATA_KEY = "crlRequestData";
+
+  /** Builds a standard {@link Context} object for calling Azure Resource Manager APIs. */
+  public static Context buildContext(ResourceManagerRequestData requestData) {
```
wchamber, line 16: probably worth saying more about when this should be used and what it does.

azure-resourcemanager-compute/src/test/java/bio/terra/cloudres/azure/resourcemanager/compute/ComputeManagerTest.java
rtitle: Yeah central test makes sense. Maybe there is a way to intercept the logger and do some validation as well. Will experiment..

common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java
rtitle: Yeah the way the Azure library is structured it would be hard/impossible to wrap everything with `executeCowOperation` while retaining all the creation info. For example a VM creation has a sequence of `DefinitionStages` that must be defined in order:

https://github.com/Azure/azure-sdk-for-java/blob/eb31e63e13764925a729ebb3b36ea0d802798e13/sdk/resourcemanager/azure-resourcemanager-compute/src/main/java/com/azure/resourcemanager/compute/models/VirtualMachine.java#L432

The `create` method is defined by a simple interface: 

https://github.com/Azure/azure-sdk-for-java/blob/eb31e63e13764925a729ebb3b36ea0d802798e13/sdk/resourcemanager/azure-resourcemanager-resources/src/main/java/com/azure/resourcemanager/resources/fluentcore/model/Creatable.java

It's all very fluent and interface driven. There are `*Impl` classes but most fields are `private` so it seems it's not really designed to be extended/wrapped. The fluent API actually makes the library very nice to work with.

Luckily the loggers seem to get the job done with minimal intervention!

azure-resourcemanager-compute/src/main/java/bio/terra/cloudres/azure/resourcemanager/compute/data/CreateDiskRequestData.java
rtitle: I'm actually leaning against this. Composition removes the need for duplicate setters, but then callers need to deal with nested objects which I feel is a little more painful..

PR #101, opened 2021-11-22
[IA-2938] Enable Janitor for Azure resources

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2938

Several Azure-related improvements:

* Updated Janitor client version to include Azure `CloudResourceUid` models
   * Since this is a common dependency change, bumped each module version and bumped `platform` version
* Wire up Janitor clean-up messages for all Azure resources (it works!)
* Changed some Azure gradle dependencies to `api` instead of `implementation` so they get pulled in as part of CRL
* Pull Azure creds from vault, enable integration test
* Fixed a couple failing GCP integration tests due to a Broad networking change (the default network is no longer persent in GCP projects)

Related: https://github.com/DataBiosphere/terra-resource-janitor/pull/120  (still need to update this PR to consume the new CRL version)

Tested by running Janitor locally listening to a pubsub topic that I created; and running a CRL integration test that creates an Azure IP. Janitor receives the message and deletes the IP!


google-compute/src/test/java/bio/terra/cloudres/google/compute/CloudComputeCowTest.java
```diff
@@ -117,9 +117,23 @@ public void createGetAndDeleteFirewall() throws Exception {
 
     CloudComputeCow cloudComputeCow = defaultCompute();
 
+    // Create parent network
```
rtitle, line 120: Not related to Azure, but these `CloudComputeCowTest`s were failing due to:
```
"The resource 'projects/pa081db47-9b33-4c3a-b43e-8f3a0/global/networks/default' was not found",
```
I think the `default` network doesn't exist anymore; I made the tests create a parent network.

yonghaoy: Thanks! Those tests are created within broadinstitute.org, and bits turn off an org policy that removes default network for the entier org

yonghaoy (approved):
LG, If you want to use Janitor, you need to publish the message to janitor, and make sure Janitor understand that. Do we already have this logic in CRL.
I saw we have code here: 
https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/azure-resourcemanager-common/src/main/java/bio/terra/cloudres/azure/resourcemanager/common/AzureResourceCleanupRecorder.java#L40L43
Does that work with Janitor already?

rtitle:
> Does that work with Janitor already?

Excellent question - I think it should work but I will test. The Janitor changes aren't deployed yet but perhaps I can verify locally.

yonghaoy:
> > Does that work with Janitor already?
> 
> Excellent question - I think it should work but I will test end-to-end. The Janitor changes aren't deployed to GKE yet but perhaps I can verify locally.

SG. Also an optional tip: You can setup your own personal GKE env then deploy your local janitor/RBS/WSM Apps.
Here is an example(that env name is buffertest, but it is actually just a personal env): 
https://github.com/broadinstitute/terraform-ap-deployments/pull/245/files
Then https://github.com/broadinstitute/terra-helmfile/pull/1117
After you have terraform and helm, you can bring up Janitor server using skaffold:
https://github.com/DataBiosphere/terra-resource-janitor#deploy-to-gke-cluster

it might be helpful if you plan to work on Janitor a lot. And I know IA is actually working on WSM changes, it would be the same process to bringup a personal WSM.

google-compute/src/test/java/bio/terra/cloudres/google/compute/CloudComputeCowTest.java
rtitle: Btw I think the notebooks tests also fail due to this. When creating a notebook instance:
```
Invalid value for field 'resource.networkInterfaces[0].network': ''. The referenced network resource cannot be found.
```
Will see if I can create a network in those tests too..

yonghaoy: Thanks. Don't need to spend too much time if that is too complicated. I did plan to move those test folders to test.fc.org

rtitle:
I tried to test end-to-end and found that CRL is not sending the pubsub messages. I think I need to update CRL to the latest janitor _client_ (`0.5.0-SNAPSHOT`) then update the Azure `*RequestData` classes to create `CloudResourceUid` objects to publish. Think this also means version bumping all CRL module versions (since a `common` dependency is changing)..

yonghaoy:
> I tried to test end-to-end and found that CRL is not sending the pubsub messages. I think I need to update CRL to the latest janitor _client_ (`0.5.0-SNAPSHOT`) then update the CRL `*RequestData` classes to create `CloudResourceUid` objects to publish. Think this also means version bumping all CRL modules (since a `common` dependency is changing)..
> 
> Feels like we have some circular dependencies here making it a bit hard to add a new resource type. 😅 But I'll give it a shot in this PR..

Yes, its bit awkward. Initially we thought it might be helpful to create a "unified way to represent cloud resources" that can be used by all Terra Apps. And CRL is the best place to hold this value. 
But managing/publishing the version of the `CloudResourceUid` is bit awkward. 
The usually steps is (1) Update `cloudResorceUid` while implementing CRL support. 
(2) Implement Janitor with the new `CloudResourceUid` version.
(3)Go back to bump Janitor version in CRL.

Now it turns out the `cloudResourceUid` is not useful and I am thinking about changing it or deprecating it. Some options I have been thinking of:
(1) Find others ways to publish cloud_resource_uid.yaml (not using artifact registry, probably GitHub Release)
(2) Deprecating `CloudResourceUid`, and move that into Janitor. 

2 is slightly preferred because `CloudResourceUid` is not really useful for now.  I owe that task for a while because AoU side work, but I can pick that task up after holiday. 

Do you have preferences on 1 or 2 or other options? Suggestions/recommendations are appreciated!

rtitle:
Ah I'm working on this late too. :) 

I basically did your suggestion and got it working end-to-end. Re-requesting reviews, I think this is mergeable now.

For the future, I think (2) is probably more intuitive -- Janitor owns the PubSub schema so it should own `CloudResourceUid` I think. We might still have some multi-step process for making updates to CRL and Janitor, but at least the yaml file and generated Java code won't be in different repos.

yonghaoy (approved):

.github/workflows/build-and-test.yaml
```diff
@@ -56,8 +59,23 @@ jobs:
...
             vault read -format json $VAULT_SERVICE_ACCOUNT_JANITOR_CLIENT_PATH \
-            | jq -r .data.key | base64 -d > $SERVICE_ACCOUNT_JANITOR_CLIENT_FILE \
-            | jq .data > SERVICE_ACCOUNT_JANITOR_CLIENT_FILE
+            | jq -r .data.key | base64 -d > $SERVICE_ACCOUNT_JANITOR_CLIENT_FILE &&#
+          docker run --rm --cap-add IPC_LOCK \
+            -e "VAULT_TOKEN=${{ steps.vault-token-step.outputs.vault-token }}" \
+            -e "VAULT_ADDR=${VAULT_ADDR}" \
+            vault:1.1.0 \
+            vault read -format json $VAULT_AZURE_MANAGED_APP_CLIENT_PATH \
+            | jq .data > $AZURE_MANAGED_APP_CLIENT_FILE
```
Qi77Qi, line 68: this probably ok too..but I think you can also use terraform to push vault secrets to github as secrets that can be used in workflows...See https://github.com/broadinstitute/terraform-ap-deployments/pull/459

Qi77Qi (approved):

.github/workflows/build-and-test.yaml
rtitle: Ah good to know -- this workflow has some existing `vault read`s so I followed the pattern.

There's probably some larger work to be done about Terraforming Azure envs/secrets in general..

PR #102, opened 2022-01-05
[no ticket] Fix typo in render-test-config.sh

rtitle:
We changed the vault info since this was originally added. This should hopefully fix integration tests.

PR #103, opened 2022-01-05
[PF-1152] Add subnetworks.aggregatedList to CloudComputeCow

melissachang:
Needed for https://github.com/DataBiosphere/terra-resource-buffer/pull/206

Tested by logging aggregatedList response, it looked right.

CI failing test will be fixed by #102

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,9 @@ Broad employees can get the credentials from Vault with:
 Then actually run the tests with:
 ```
 ./gradlew integrationTest
+
+# Run a single test
+/gradlew :google-bigquery:integrationTest --tests BigQueryCowTest.listDataset
 ```
 
 TODO: Make this easier, consider making credentials path configurable.
--- a/README.md
+++ b/README.md
@@ -89,9 +89,11 @@ Broad employees can get the credentials from Vault with:
 Then actually run the tests with:
 ```
 ./gradlew integrationTest
+```
 
-# Run a single test
-/gradlew :google-bigquery:integrationTest --tests BigQueryCowTest.listDataset
+Run a single test:
+```
+./gradlew :google-bigquery:integrationTest --tests BigQueryCowTest.listDataset
 ```
 
 TODO: Make this easier, consider making credentials path configurable.
--- a/google-compute/gradle.properties
+++ b/google-compute/gradle.properties
@@ -1,3 +1,3 @@
 # Please update platform to consume version updates, see here for more:
 # https://github.com/DataBiosphere/terra-cloud-resource-lib#publishing-an-update
-version = 0.14.0
+version = 0.13.1
--- a/google-compute/gradle.properties
+++ b/google-compute/gradle.properties
@@ -1,3 +1,3 @@
 # Please update platform to consume version updates, see here for more:
 # https://github.com/DataBiosphere/terra-cloud-resource-lib#publishing-an-update
-version = 0.13.0
+version = 0.14.0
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -89,6 +89,9 @@ Broad employees can get the credentials from Vault with:
 Then actually run the tests with:
 ```
 ./gradlew integrationTest
+
```
yonghaoy, line 92: nit: missing two ```

Screen Shot 2022-01-05 at 3 06 00 PM

platform/build.gradle
```diff
@@ -14,7 +14,7 @@ dependencies {
     api group: 'bio.terra.cloud-resource-lib', name: 'google-bigquery', version: '0.12.0'
     api group: 'bio.terra.cloud-resource-lib', name: 'google-billing', version: '0.11.0'
     api group: 'bio.terra.cloud-resource-lib', name: 'google-cloudresourcemanager', version: '1.4.0'
-    api group: 'bio.terra.cloud-resource-lib', name: 'google-compute', version: '0.13.0'
```
jmczerk, line 17: I think your original change to `0.14.0` was correct, both per our general usage (I don't think we generally use patch versions) and semantic versioning conventions (since we added a new API, that would generally call for a minor bump in semver).

Platform version should also bump to `0.5.0`, per your original commit.  

I believe RB does use the platform: https://github.com/DataBiosphere/terra-resource-buffer/blob/master/build.gradle#L79

It just requires a workaround for the `cloud-resourece-schema` library specifically: https://github.com/DataBiosphere/terra-resource-buffer/blob/master/build.gradle#L47

google-compute/gradle.properties
```diff
@@ -1,3 +1,3 @@
 # Please update platform to consume version updates, see here for more:
 # https://github.com/DataBiosphere/terra-cloud-resource-lib#publishing-an-update
-version = 0.13.0
```
yonghaoy, line 3: The previous version bump pattern is 0.13 to 0.14, we never 0.13.1 before.

google-compute/src/test/java/bio/terra/cloudres/google/compute/CloudComputeCowTest.java
```diff
@@ -23,6 +24,8 @@
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
yonghaoy, line 28: those two are not used. Did `./gradlew spotlessApply` remove them?

yonghaoy (approved):

jmczerk (approved):

google-compute/src/test/java/bio/terra/cloudres/google/compute/CloudComputeCowTest.java
melissachang: Done, thanks. In gradle file, I changed `build.dependsOn spotlessApply` to `compileJava.dependsOn spotlessApply`. This makes it so that when I run the test, spotless is run. Most repos have this (the latter).

melissachang: Changing `build.dependsOn spotlessApply` to `compileJava.dependsOn spotlessApply` caused some things to fail. I reverted that change. I'll redo in a separate PR.

PR #104, opened 2022-01-06
Fix flakiness in BigQueryCowTest.listDataset()

melissachang:
Before this PR, BigQueryCowTest.listDataset() failed 4-10 times out of 10. After this PR, passes 10/10.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -11,12 +11,14 @@ plugins {
   id 'groovy-gradle-plugin'
 }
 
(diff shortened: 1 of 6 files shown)


yonghaoy (approved):
Thanks. Since this changes CRL BQ library code, we may also want to bump that package version.

melissachang:
> Thanks. Since this changes CRL BQ library code, we may also want to bump that package version.

Good point, done.

jmczerk (approved):

PR #106, opened 2022-01-13
[no ticket] Clean up Azure network security groups in Janitor

rtitle:
I noticed that Azure network security groups were not getting cleaned up by Janitor. I think this is because CRL was not recording the `CloudResourceUid` for NSG creation (it's working for the rest of the Azure resources).

I added some unit tests to catch this; and also updated Azure library versions while I'm in here.

Tested by publishing locally and running WSM `./gradlew azureTest` with the updated CRL version.


rtitle:
I don't know why `IamCowTest` is consistently failing 😢

google-iam/src/test/java/bio/terra/cloudres/google/iam/IamCowTest.java
```diff
@@ -63,18 +65,23 @@ public void createAndGetAndListAndDeleteServiceAccount() throws Exception {
             .build();
     // Retry a lot of times. It apparently takes a lot of time for GCP to be ready to get a service
     // account that was just created.
+    ServiceAccount retrievedServiceAccount = null;
```
rtitle, line 68: This test is unrelated to Azure, but it was consistently failing in my PR. Retrying SA `get` and `list` seems to fix it.

yonghaoy (approved):

Qi77Qi (approved):
thank you so much @rtitle !

PR #109, opened 2022-02-09
add CloudBillingClient#testIamPermissions

rtitle:
Needed for Billing Profile Manager.

### DIFF
--- a/google-billing/gradle.properties
+++ b/google-billing/gradle.properties
@@ -1,3 +1,3 @@
 # Please update platform to consume version updates, see here for more:
 # https://github.com/DataBiosphere/terra-cloud-resource-lib#publishing-an-update
-version = 0.11.1
+version = 0.12.0
--- a/google-billing/src/main/java/bio/terra/cloudres/google/billing/CloudBillingClientCow.java
+++ b/google-billing/src/main/java/bio/terra/cloudres/google/billing/CloudBillingClientCow.java
@@ -9,7 +9,10 @@ import com.google.cloud.billing.v1.CloudBillingSettings;
 import com.google.cloud.billing.v1.ProjectBillingInfo;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.gson.Gson;
+import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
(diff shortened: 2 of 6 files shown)


cahrens:
@rtitle I ran across a TODO in BPM about this PR. Why was it never merged? https://github.com/DataBiosphere/terra-billing-profile-manager/blob/ef8f6359ce3d1ef68cef575ea0b07d8c378dbe0f/service/build.gradle#L41

rtitle:
Um, it's hazy but I think I was experiencing connected test failures in CRL around project allocation. (Of course, the test logs are gone now.) Also I can't spot any particular issues at the moment, but I have a recollection that GCP support in BPM needed a bit more testing and verification of correctness -- I may have been holding off merging this until we confirmed it with BPM.

cahrens:
Some of these files don't even exist anymore. I'm concerned that BPM appears stuck on this very old version.

rtitle:
CRL at some point after this PR underwent a large refactor to un-modularize things -- there used to be a gradle module for each cloud library (google-billing, azure-compute, etc). Now it's just 1 dependency (with packages per cloud library).

cahrens:
So… does that mean that BPM should change how its dependency is declared? But presumably there was something important in this PR that was needed for BPM, so I suspect this would cause something to break? https://github.com/DataBiosphere/terra-billing-profile-manager/blob/ef8f6359ce3d1ef68cef575ea0b07d8c378dbe0f/service/build.gradle#L41

rtitle:
Re how the dependency is declared -- yes I think so.

I think BPM uses the method added in this PR [here](https://github.com/DataBiosphere/terra-billing-profile-manager/blob/main/service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileVerifyAccountStep.java#L35)

Sorry for leaving this not finished.. probably the right next steps are to get BPM on the latest CRL, and either re-do this functionality, or temporarily remove it from `CreateProfileVerifyAccountStep` (since BPM is not yet being used with GCP anyway).

cahrens:
OK, I created https://broadworkbench.atlassian.net/browse/WOR-714

PR #113, opened 2022-02-16
Update cloud notebook version

yuhuyoyo:
the rest of the changes are from running `./minnie-kenny.sh -f` to fix test failure at master branch.

PR #115, opened 2022-02-17
bump cloud notebook version to 1.0

yuhuyoyo:
code changes is here: https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/113/files#diff-37ff50b008a551eb713f147269736de3233c74488bd6acc324cb64120d51aaec

### DIFF
--- a/google-notebooks/gradle.properties
+++ b/google-notebooks/gradle.properties
@@ -1,3 +1,3 @@
 # Please update platform to consume version updates, see here for more:
 # https://github.com/DataBiosphere/terra-cloud-resource-lib#publishing-an-update
-version = 1.0.0
+version = 0.10.0
--- a/google-notebooks/gradle.properties
+++ b/google-notebooks/gradle.properties
@@ -1,3 +1,3 @@
 # Please update platform to consume version updates, see here for more:
 # https://github.com/DataBiosphere/terra-cloud-resource-lib#publishing-an-update
-version = 0.9.0
+version = 1.0.0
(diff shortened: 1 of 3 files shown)


google-notebooks/gradle.properties
```diff
@@ -1,3 +1,3 @@
 # Please update platform to consume version updates, see here for more:
 # https://github.com/DataBiosphere/terra-cloud-resource-lib#publishing-an-update
-version = 0.9.0
```
jmczerk, line 3: I don't think we want to go to major rev 1 yet, so I think this should be `0.10.0`.

jmczerk: You should also change the platform to consume this new version here:
https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/platform/build.gradle#L20

And bump the platform version to `0.9.0` here:
https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/platform/gradle.properties

jmczerk: Note how WSM uses the platform here rather than individual library versions:

https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/build.gradle#L50

jmczerk: Yes, any downstream library that uses the CRL notebook library needs to update their dependency version in order to consume the change you made.  You will need to wait until the new artifacts are published to make that change downstream (i.e. step 3 [here](https://github.com/DataBiosphere/terra-cloud-resource-lib#publishing-an-update)).  

As the docs say, that job will likely fail, but the new artifacts will likely publish anyway.

jmczerk (approved):

Qi77Qi (approved):

PR #119, opened 2022-03-15
[PF-722] Require manual publish, stating artifact names

jaycarlton:
To avoid the problem of re-publishing already-published artifact versions (and failing due to overwrite protection), introduce a string input and `workflow_dispatch` trigger to take all and only the artifacts to be published. It's expected that `platform` will be included every time (though not currently enforced).


jaycarlton:
Ended up going a different direction.

PR #123, opened 2022-03-28
[PF-722] fix branch & add local action

jaycarlton:
`master` the correct default checkout branch. We also need to have a copy of `bump-skip` from `terra-workspace-manager` in the local `.github/actions` directory.

Created [PF-1431](https://broadworkbench.atlassian.net/browse/PF-1431) to publish the action somewhere appropriate so we don't maintain two or more copies.


zloery (approved):
+1 to centralizing this action eventually, but LGTM for now

melissachang (approved):

PR #129, opened 2022-04-11
[PF-1531] Update dependencies, use new lockfile format

zloery:
This bumps the version of several dependencies with identified CVEs, as we rely on these in downstream services. The main changes are:

com.google.cloud:libraries-bom 20.2.0 -> 25.1.0
com.google.guava:guava 30.1.1-jre -> 31.1-jre

This also refactors the lockfiles into a single file. This is the new automatic behavior from `dependencies --write-locks` after the gradle update in #128.


yuhuyoyo (approved):

src/test/java/bio/terra/cloudres/google/billing/CloudBillingClientCowTest.java
```diff
@@ -64,7 +64,7 @@ public void serialize() {
         "{\"project_name\":\"projects/my-project\","
             + "\"project_billing_info\":{\"name_\":\"\",\"projectId_\":\"my-project\","
             + "\"billingAccountName_\":\"billingAccounts/01A82E-CA8A14-367457\",\"billingEnabled_\":false,"
-            + "\"memoizedIsInitialized\":1,\"unknownFields\":{\"fields\":{},\"fieldsDescending\":{}},"
+            + "\"memoizedIsInitialized\":1,\"unknownFields\":{\"fields\":{}},"
```
jaycarlton, line 67: Was this change intended for this PR?

jaycarlton (approved):

src/test/java/bio/terra/cloudres/google/billing/CloudBillingClientCowTest.java
zloery: Yes, this is a real change caused by the cloud library version bump (actual cause is a transitive change in the protobuf library). I don't think there are compatibility issues given the way CRL uses serialized requests

PR #130, opened 2022-04-11
[PF-1531] Bump Artifactory plugin version to (hopefully) fix publishing

zloery:
My previous PR failed to publish. The error from the build scan points to the Artifactory gradle plugin being unhappy and this started showing up with the move to gradle 7.4, so hopefully this will stop the action from failing.


zloery:
Also: it looks like the version bumper is really confused by this repo's mix of tags using/not using the `v` prefix ([see downgrade bump here](https://github.com/DataBiosphere/terra-cloud-resource-lib/commit/73144e0a66bd611b5e48ca80cbddc71364896fed)), which is causing problems for keeping track of versions. Not sure if it makes more sense to go fix the action upstream or just start using `v` prefixes here.

jaycarlton:
This might not be you, since we were having trouble with it before. I was able to manually publish v1.1.0, but we were waiting for the next commit to see if it worked.

jaycarlton (approved):

zloery:
I did a bit more digging and I think it's a problem with the `bumper` action, it assumes any version with a `v` prefix is higher than any version without it. In this repo it keeps thinking `v1.0.3` is the latest and tries to publish `1.0.4` (no prefix), so we end up stuck in a loop. I opened a PR to fix it (https://github.com/DataBiosphere/github-actions/pull/31) so once that's in I can pull that in here.

Alternatively we can just manually publish `v1.1.1` and start using tags with the `v` prefix (which is a flag in the bump-tag workflow).

jaycarlton:
Great find! The tags in this repo are a hot mess, so it's good to know there's a method to the madness.

I'm fine waiting for the fix to `bumper`, since it might affect other repos too. Is this a bug in an upstream `actions/bumper` as well?

yuhuyoyo (approved):

PR #135, opened 2022-07-10
[TOAZ 100] Landing Zone manager and documentation

giventocode:
- This PR includes the first implementation of the LZ library.
- For more details, check the [docs/landingzones/README.md](./docs/landingzones/README.md)


src/main/java/bio/terra/cloudres/azure/landingzones/definition/factories/LandingZoneDefinitionProviderImpl.java
```diff
@@ -0,0 +1,76 @@
...
+          .map(c -> toFactoryInfo((Class<? extends LandingZoneDefinitionFactory>) c.load()))
+          .collect(Collectors.toSet());
+    } catch (IOException e) {
+      throw logger.logExceptionAsError(new RuntimeException(e));
+    }
+  }
+
+  private boolean isLandingZoneFactory(ClassInfo classInfo) {
+    // a factory is a non-abstract class that implements LandingZoneDefinitionFactory.
+    return !classInfo.load().isInterface()
```
jordansafer, line 44: Is there a simpler way to check that it's a LZFactory? Can we use `isinstance` to check if it's an instance of a landingZoneFactory?

src/main/java/bio/terra/cloudres/azure/landingzones/definition/factories/ManagedNetworkWithSharedResourcesFactory.java
```diff
@@ -0,0 +1,108 @@
...
+    @Override
+    public Deployable definition(DefinitionContext definitionContext) {
+      WithLandingZoneResource deployment = definitionContext.deployment();
+      ResourceGroup resourceGroup = definitionContext.resourceGroup();
+      ResourceNameGenerator nameGenerator = definitionContext.resourceNameGenerator();
+
+      var storage =
+          azureResourceManager
+              .storageAccounts()
+              .define(nameGenerator.nextName(23))
```
jordansafer, line 66: similar to above, why is the name length 23 here?

src/main/java/bio/terra/cloudres/azure/landingzones/deployment/LandingZoneDeploymentImpl.java
```diff
@@ -0,0 +1,122 @@
...
+  private final ResourcesTagMapWrapper resourcesTagMapWrapper;
+
+  protected LandingZoneDeploymentImpl(ResourcesTagMapWrapper resourcesTagMapWrapper) {
+    this.resourcesTagMapWrapper = resourcesTagMapWrapper;
+  }
+
+  @Override
+  public List<DeployedResource> deploy() {
+
+    return deployAsync().collectList().block();
```
jordansafer, line 29: Do we need a separate deployAsync() and deploy() method in this case? I think the sync caller can block() the async method-

src/main/java/bio/terra/cloudres/azure/landingzones/deployment/ResourcesTagMapWrapper.java
```diff
@@ -0,0 +1,104 @@
...
+
+  <T extends Creatable<?> & Resource.DefinitionWithTags<?>> void putWithPurpose(
+      T resource, ResourcePurpose purpose) {
+
+    putWithLandingZone(resource);
+    putTagKeyValue(
+        resource, LandingZoneTagKeys.LANDING_ZONE_PURPOSE.toString(), purpose.toString());
+  }
+
+  void putResource(Creatable<?> resource) {
```
jordansafer, line 54: It looks like this method overwrites the tagMap for a resource. Could we call it "resetResource"?

src/main/java/bio/terra/cloudres/azure/landingzones/management/LandingZoneManager.java
```diff
@@ -0,0 +1,135 @@
...
+        new ResourceNameGenerator(landingZoneId),
+        new HashMap<>());
+  }
+
+  public ResourcesReader reader() {
+    return resourcesReader;
+  }
+
+  public LandingZoneDeployments deployments() {
+    return landingZoneDeployments;
```
jordansafer, line 129: When we call `deployLandingZoneAsync`, how does the new deployment get added to the list of landingZoneDeployments? I'm not sure I follow this part-

jordansafer (approved):
2 PR level comments:
- We have custom methods/logic paths for VNet/Subnet resources. Is it worth avoiding this (most likely by adding wrapper types)? Are there other resources beyond VNet which will require additional logic in this way for managing/deploying them?
- How will we store/persist landing zones? Of course this will come outside of the library. Do we want to store the definitionContext information (is this enough information to fully identify the landing zone? or too much information)?

src/main/java/bio/terra/cloudres/azure/landingzones/definition/ResourceNameGenerator.java
```diff
@@ -0,0 +1,80 @@
...
+    end = Math.min(end, hash.length() + 2);
+    return "lz" + hash.substring(0, end - 2);
+  }
+
+  private String parseStringToHash() {
+    return String.format("%s%s", landingZoneId, sequence);
+  }
+
+  public String nextName() {
+    return nextName(66);
```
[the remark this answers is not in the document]
giventocode: This is the max length of the name. I will add a constant to make it more explicit.

src/main/java/bio/terra/cloudres/azure/landingzones/definition/factories/LandingZoneDefinitionProviderImpl.java
giventocode: The intent here is to check if a class in a package is a factory, and not whether an instance is. The library uses reflection to locate all factories in the factory package. And this check filters out, abstract classes and the interfaces, in short, only factories that can be instantiated are considered valid factories. This is how the library lists the available factories.

src/main/java/bio/terra/cloudres/azure/landingzones/deployment/LandingZoneDeploymentImpl.java
giventocode: I believe having two async and not async options at the library level is okay.

src/main/java/bio/terra/cloudres/azure/landingzones/deployment/ResourcesTagMapWrapper.java
giventocode: The method has the same semantics as the `put` method in a map -put overwrites an existing key value. So, I think its name is appropriate

src/main/java/bio/terra/cloudres/azure/landingzones/management/LandingZoneManager.java
giventocode: deployments are not added to the list because there's no current way to persist them - this is something we can add later. The intent here is to have a fluent interface, at the high level, where the `lzManager.deployments().define(lzid)` is a fluent builder of deployments. This is the same pattern as the Azure Java SDK, e.g. `arm.storageAccounts().define(name)... `

src/main/java/bio/terra/cloudres/azure/landingzones/definition/DefinitionVersion.java
```diff
@@ -0,0 +1,22 @@
...
+
+import com.azure.core.util.ExpandableStringEnum;
+
+/** Enum representing the list of possible versions for a given definition. */
+public final class DefinitionVersion extends ExpandableStringEnum<DefinitionVersion> {
+  public static final DefinitionVersion V1 = fromString("v1");
+  public static final DefinitionVersion V2 = fromString("v2");
+  public static final DefinitionVersion V3 = fromString("v3");
+  public static final DefinitionVersion V4 = fromString("v4");
+  public static final DefinitionVersion V5 = fromString("v5");
```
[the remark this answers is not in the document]
giventocode: Each definition indicates what versions are available (a subset of these), so having an enum type with more values seemed harmless, and better than having an enum with a single value.

giventocode:
> 2 PR level comments:
> 
> * We have custom methods/logic paths for VNet/Subnet resources. Is it worth avoiding this (most likely by adding wrapper types)? Are there other resources beyond VNet which will require additional logic in this way for managing/deploying them?

VNEts are a special case because the purpose is actually associated with the subnet and they can have multiple subnets. Resources that can't be tagged also have a special with method - `withResource`.  
 
> * How will we store/persist landing zones? Of course this will come outside of the library. Do we want to store the definitionContext information (is this enough information to fully identify the landing zone? or too much information)?

The landing zone ID plus the MRG and the subscription should be enough to locate the landing zone.

rtitle:
Question - should this PR be closed now that the functionality has moved to https://github.com/DataBiosphere/terra-landing-zone-service/tree/main/service/src/main/java/bio/terra/landingzone/library?

PR #145, opened 2022-09-28
Bump azure-core version to be bring in jdk 17 fix

aherbst-broad:
We are bumping into [this bug](https://github.com/Azure/azure-sdk-for-java/pull/30195) in Sam. This is fixed in azure-core, so I'm bumping the version here so we can test the fix in Sam. I was looking at shifting to using the azure-sdk-bom for easier version management but that does not include the resourcemanager libs.


rtitle (approved):
LGTM, if it fixes the issue in Sam

marctalbott (approved):

PR #146, opened 2022-10-03
[PF-1543] CRL: Java version upgrade to 17

dexamundsen:
- Java version upgrade to 17
- Update Build files to keep them consistent with other terra repositories
- Upgrade dependency versions


gradle/test.gradle
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// Create a new :integrationTest task to run integration tests. This won't be executed as
+// part of the standard `./gradlew build` invocation.
+task integrationTest(type: Test) {
+    useJUnitPlatform {
+        includeTags "integration"
+    }
+    shouldRunAfter test
+    // Force tests to always be re-run, since integration tests involve communicating with external
```
yuhuyoyo, line 26: how does upToDateWhen force re-run? i thought you need to have a retry clause like this 
https://cs.github.com/DataBiosphere/terra-workspace-manager/blob/f3b3891897879b203c81795c7d322c2a9df93091/service/gradle/testing.gradle?q=connectedtest#L56

gradle/dependency-locking.gradle
```diff
@@ -0,0 +1,3 @@
+dependencyLocking {
+    lockAllConfigurations() // see https://docs.gradle.org/current/userguide/dependency_locking.html
```
yuhuyoyo, line 2: thanks for splitting them up!

yuhuyoyo (approved):

gradle/test.gradle
dexamundsen: Agree, Only split the build files in this PR, but retained the task config from before. This needs to be updated
Added a TODO comment

.github/workflows/build-and-test.yml
```diff
@@ -6,8 +6,11 @@ name: Java CI with Gradle
 on:
   push:
     branches: [ master ]
+    paths-ignore: [ '**.md' ]
```
zloery, line 9: is this workflow require to merge PRs? if so we should still run on .md file changes so they don't get stuck without a required check

zloery (approved):

.github/workflows/build-and-test.yml
dexamundsen: Nope,

'push' trigger is invoked after the change is merged to master
'Pull_request' trigger is used for PRs and hence wont block PRs

PR #147, opened 2022-11-02
[PF-2181] Add Compute.Zones.List

iliadsh:
https://cloud.google.com/compute/docs/reference/rest/v1/zones/list

### DIFF
--- a/src/main/java/bio/terra/cloudres/google/compute/CloudComputeCow.java
+++ b/src/main/java/bio/terra/cloudres/google/compute/CloudComputeCow.java
@@ -619,6 +619,61 @@ public class CloudComputeCow {
         return result;
       }
     }
+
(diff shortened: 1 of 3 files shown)


melissachang (approved):
nit we usually put ticket name in PR title

```
[PF-2162] Add Compute.Zones.List
```

This is useful in both directions:

- If someone's looking at code, they can go to Jira ticket for more context
- If someone's looking at Jira ticket, they can see PRs. Example:

image

"See PRs in Jira" doesn't work for https://github.com/verily-src/terra-ui because it's a private repo. We use Broad's Jira instance; it doesn't have access to verily-src. Eventually we'll use the Verily Jira instance, which will have access to verily-src.

This is a public repo so "See PRs in Jira" will work.

cbookg (approved):

PR #151, opened 2022-12-16
[TERRA-225] Add notebooks.cow for AWS + rebase onto master

dexamundsen:
- Add notebooks.cow for AWS, similar to the ones used by GCP
- Rebase onto master 

Tested with minor object changes in
https://github.com/DataBiosphere/terra-cli/pull/375


dexamundsen:
Discussed with JohnC: This code will be retained in CLI until the time it needs to be moved to CRL for shared usage

PR #158, opened 2023-04-04
[TERRA-384] Add basic AWS endpoint wrappers

zloery:
This wraps basic SageMaker notebook and S3Bucket endpoints to CRL in order to log requests and record metrics for these calls.

Because these are generally thin wrappers, this is tested entirely via unit tests which do not talk to AWS infrastructure. Services using this library should write integration/connected tests as needed to verify AWS behavior.


src/main/java/bio/terra/cloudres/aws/bucket/S3BucketCow.java
```diff
@@ -0,0 +1,133 @@
...
+            S3BucketOperation.AWS_GET_S3_OBJECT,
+            () ->
+                bucketClient.getObject(
+                    GetObjectRequest.builder().bucket(bucketName).key(objectPath).build()),
+            () -> serialize(bucketName, objectPath))) {
+      return response.response();
+    }
+  }
+
+  public ListObjectsV2Response listBlobs(String bucketName, String prefix) {
```
jmczerk, line 97: The `ListObjectV2` API has a limit of 1,000 object returned per call.  We'll want to plumb the ability to optionally pass a [continuation token](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/model/ListObjectsV2Request.Builder.html#continuationToken(java.lang.String)) in case the response comes back as truncated and containing a continuation token, in case we need to iterate more than 1,000 object.

src/main/java/bio/terra/cloudres/aws/notebook/SageMakerNotebookCow.java
```diff
@@ -0,0 +1,194 @@
...
+  public void stop(String instanceName) {
+    operationAnnotator.executeCowOperation(
+        SageMakerNotebookOperation.AWS_STOP_NOTEBOOK,
+        () ->
+            notebooksClient.stopNotebookInstance(
+                StopNotebookInstanceRequest.builder().notebookInstanceName(instanceName).build()),
+        () -> createJsonObjectWithSingleField("instanceName", instanceName));
+  }
+
+  /**
```
jmczerk, line 128: `s/start/stop`
`s/usable/stopped`

jmczerk (changes requested):
One change requested, but otherwise looks great.  Thanks for taking this on!

jmczerk (approved):

dexamundsen (approved):

PR #166, opened 2023-04-19
[TERRA-498] Add folderExists and deleteFolder to s3BucketCow, surface errors

zloery:
This is a small collection of improvements to the S3BucketCow as we've iterated on adding S3 buckets to WSM. The main changes:

- Add unchecked exceptions from the underlying S3 client to method signatures. This makes it clearer what exceptions these methods will throw in common failure modes
- Add a `folderExists` method as a wrapper around `listBlobs`
- Add a `deleteFolder` method which deletes multiple objects in a single call. This uses a different AWS endpoint, so I'm logging it as a different operation.
- Allow callers to pass tags in during object creation
- Fix `assertEquals` in tests, the expected and actual arguments were swapped.

In addition to the CRL tests, see [example WSM PR](https://github.com/DataBiosphere/terra-workspace-manager/pull/1210) for a preview of how these methods can be used in WSM.

### DIFF
--- a/src/main/java/bio/terra/cloudres/aws/bucket/S3BucketCow.java
+++ b/src/main/java/bio/terra/cloudres/aws/bucket/S3BucketCow.java
@@ -106,9 +106,9 @@ public class S3BucketCow implements AutoCloseable {
   }
 
   /**
-   * Delete all objects in an AWS bucket with a common prefix. Because AWS can only support up to
-   * 1000 object deletions from a single request, this method may make multiple calls to AWS
-   * infrastructure, each of which will be logged separately.
+   * Delete all objects in an AWS bucket with a common prefix, including the folder itself. Because
+   * AWS can only support up to 1000 object deletions from a single request, this method may make
+   * multiple calls to AWS infrastructure, each of which will be logged separately.
    */
   public void deleteFolder(String bucketName, String prefix)
       throws AwsServiceException, SdkClientException, S3Exception {
@@ -128,7 +128,7 @@ public class S3BucketCow implements AutoCloseable {
                               .bucket(bucketName)
                               .delete(Delete.builder().objects(partitionedObjectList).build())
                               .build()),
-                  () -> serialize(bucketName, folderKey));
+                  () -> serialize(bucketName, folderKey, partitionedObjectList.size()));
             });
   }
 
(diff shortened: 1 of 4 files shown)


src/main/java/bio/terra/cloudres/aws/bucket/S3BucketCow.java
```diff
@@ -81,7 +105,35 @@ public void deleteBlob(String bucketName, String objectPath) {
...
+            partitionedObjectList -> {
+              operationAnnotator.executeCowOperation(
+                  S3BucketOperation.AWS_DELETE_S3_FOLDER,
+                  () ->
+                      bucketClient.deleteObjects(
+                          DeleteObjectsRequest.builder()
+                              .bucket(bucketName)
+                              .delete(Delete.builder().objects(partitionedObjectList).build())
+                              .build()),
+                  () -> serialize(bucketName, folderKey));
```
jmczerk, line 131: Could we serialize the number of objects deleted on each iteration?

src/main/java/bio/terra/cloudres/aws/bucket/S3BucketCow.java
```diff
@@ -81,7 +105,35 @@ public void deleteBlob(String bucketName, String objectPath) {
         () -> serialize(bucketName, objectPath));
   }
 
-  public GetObjectResponse getBlob(String bucketName, String objectPath) throws IOException {
+  /**
+   * Delete all objects in an AWS bucket with a common prefix. Because AWS can only support up to
```
jmczerk, line 109: I assume the folder itself is included in the `getBlobs` call, but might be worth mentioning in the comments.

src/test/java/bio/terra/cloudres/aws/bucket/S3BucketCowTest.java
```diff
@@ -157,4 +191,30 @@ public void listBlobsWithContinuationTest() {
...
+    JsonObject json = gsonArgumentCaptor.getValue();
+    JsonObject serializedRequest =
+        bucketCow.serialize(fakeBucketName, missingFolderPrefix, (String) null);
+    assertEquals(json.getAsJsonObject("requestData"), serializedRequest);
+    assertEquals(
+        json.get("operation").getAsString(), S3BucketOperation.AWS_LIST_S3_OBJECTS.toString());
+
+    when(mockS3Client.listObjectsV2((ListObjectsV2Request) any())).thenReturn(responseWithObject);
+    assertTrue(bucketCow.folderExists(fakeBucketName, realFolderPrefix));
+  }
```
jmczerk, line 219: Might be worth a case to hit the missing trailing `/` cases.

jmczerk (approved):

src/main/java/bio/terra/cloudres/aws/bucket/S3BucketCow.java
```diff
@@ -94,13 +146,14 @@ public GetObjectResponse getBlob(String bucketName, String objectPath) throws IO
...
-  public ListObjectsV2Response listBlobs(String bucketName, String prefix) {
+  public ListObjectsV2Response listBlobs(String bucketName, String prefix)
+      throws AwsServiceException, SdkClientException, S3Exception {
     // A null continuationToken is ignored by the AWS client
     return listBlobs(bucketName, prefix, /*continuationToken=*/ null);
   }
 
-  public ListObjectsV2Response listBlobs(
-      String bucketName, String prefix, String continuationToken) {
+  public ListObjectsV2Response listBlobs(String bucketName, String prefix, String continuationToken)
```
dexamundsen, line 155: We might want need to limit the number of results returned

reason - 

- Current GCP implementation in CLI limits the numObjects check to 1000 (?)
- For checkFolderExists all we need is a check for 1 prefix (the folder itself). so fetching all prefixes might be service-heavy and not needed

zloery: good call - given that we're now exposing multiple arguments to the underlying client, I've also a method taking a `ListObjectsV2Request` directly in case there are other arguments/combinations to add in the future

dexamundsen (approved):

PR #167, opened 2023-05-11
[TERRA-507] Add AWS Security Token Service COW

zloery:
This wraps the `assumeRoleWithWebIdentity` endpoint as well as a method for creating a  `AssumeRoleWithWebIdentityRequest` object. 

STS is unique in that when using an `StsAssumeRoleWithWebIdentityCredentialsProvider`, we cannot log each cloud call, as the library will automatically handle caching and refreshing credentials. Instead, this method will just log once when the initial refresh request is created which will provide some visibility, though we won't be able to track errors as easily.

### DIFF
--- a/src/main/java/bio/terra/cloudres/aws/sts/SecurityTokenServiceCow.java
+++ b/src/main/java/bio/terra/cloudres/aws/sts/SecurityTokenServiceCow.java
@@ -1,0 +1,128 @@
+package bio.terra.cloudres.aws.sts;
+
+import bio.terra.cloudres.common.ClientConfig;
+import bio.terra.cloudres.common.OperationAnnotator;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.gson.JsonObject;
+import java.time.Duration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.arns.Arn;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.sts.StsClient;
+import software.amazon.awssdk.services.sts.model.AssumeRoleRequest;
+import software.amazon.awssdk.services.sts.model.AssumeRoleWithWebIdentityRequest;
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/cloudres/aws/sts/SecurityTokenServiceCow.java
```diff
@@ -0,0 +1,128 @@
...
+   *     as the duration of a console session created from these credentials - that can be specified
+   *     separately at session creation time.
+   * @param serviceAccountEmail Email of the GCP service account used to assume this role. This is
+   *     used for the human-readable role session name.
+   * @param idToken An OIDC token for the GCP service account used to assume this role via web
+   *     identity federation.
+   * @return A request body to assume the specified role via web identity federation. Note this
+   *     method does not make any calls to STS.
+   */
+  public AssumeRoleWithWebIdentityRequest createRefreshRequest(
```
jmczerk, line 65: So I think that this might need to take a lambda/functional type that returns an ID Token string instead of parameter `String idToken`, because if a point-in-time ID token retrieved from GCP is used in the refresh request, it will eventually go stale.

Note here how we're passing a lambda to `refreshRequest` here rather than the result of the call:
https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/common/utils/AwsUtils.java#L199

zloery: Good point! Looking at this more, I'm also not sure we'd ever need a refresh request aside from building the credentials provider. Refactored this to wrap/log `createAssumeRoleCredentialsProvider` instead and to take a supplier for the token

gradle/testing.gradle
```diff
@@ -35,4 +35,4 @@ task minniekenny(type: Exec) {
     commandLine 'sh', file('minnie-kenny.sh')
 }
 
-test.dependsOn minniekenny
+//test.dependsOn minniekenny
```
jmczerk, line 38: Was this meant to stay in?

jmczerk (approved):

dexamundsen (approved):

PR #174, opened 2023-07-05
[PF2868] Add autodelete to test clusters

zloery:
This configures the test dataproc clusters to automatically delete after 30m. This should help for when tests fail to clean up their clusters. Thanks Roger for thinking of this!

This also adds a `delete` call to the one test which created a cluster but did not delete it.


rogerwangcs (approved):
Thanks for the fix zach! Probably should have had this to begin with.

yonghaoy (approved):
Usually for a new resource that is added into CRL, we add Janitor support to make sure they are deleted afterward.

PR #179, opened 2023-11-13
WOR-1337 upgrade google cloud libs

dvoet:
In the context of upgrading RBS to spring boot 3, something is pulling in later versions of google cloud libs. The storage COW serializes request data using gson but the underlying implementations of the request objects changed to use java.time objects which are not automatically serializable by gson. This PR registers some handlers for gson.


src/test/java/bio/terra/cloudres/google/storage/SerializeUtilsTest.java
```diff
@@ -67,9 +67,9 @@ public void bucketNameAndBucketTargetOption() {
     JsonObject jsonObject =
         SerializeUtils.convert(
             BucketInfo.newBuilder("my-name").build(),
-            Storage.BucketTargetOption.metagenerationMatch());
+            Storage.BucketTargetOption.userProject("my-project"));
```
dvoet, line 70: `metagenerationMatch` now produces an object that serializes to empty (i.e there is no data) so I used something else instead.

zloery (approved):
LGTM pending test issues, I think you need to rewrite the gradle lockfile as well

jmczerk (approved):
Same as @zloery , approving pending clean test run

PR #180, opened 2023-11-14
upgrade to open telemetry

dvoet:


src/main/java/bio/terra/cloudres/azure/resourcemanager/common/Defaults.java
```diff
@@ -41,7 +44,13 @@ private Defaults() {}
    * </pre>
    */
   public static Context buildContext(ResourceManagerRequestData requestData) {
-    return new Context(CLOUD_RESOURCE_REQUEST_DATA_KEY, requestData);
+    return Context.of(
+        Map.of(
+            CLOUD_RESOURCE_REQUEST_DATA_KEY,
+            requestData,
+            // azure has built-in handling of tracing, nice!
```
rtitle, line 51: 🎉

src/main/java/bio/terra/cloudres/common/ClientConfig.java
```diff
@@ -52,8 +72,22 @@ public Builder setCleanupConfig(CleanupConfig cleanupConfig) {
       return this;
     }
 
+    public Builder setOpenTelemetry(OpenTelemetry openTelemetry) {
+      this.openTelemetry = openTelemetry;
+      return this;
+    }
+
+    public Builder setMetricsHelper(MetricsHelper metricsHelper) {
```
rtitle, line 80: Does `MetricsHelper` need to be in the builder? Would any caller want to set their own, or would we just want to always construct a `new MetricsHelper(openTelemetry)`?

rtitle (approved):

zloery (approved):

src/main/java/bio/terra/cloudres/common/ClientConfig.java
dvoet: it is for testing

PR #183, opened 2024-11-19
[CORE-117] Fix google billing GSON serialization

samanehsan:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-117

**Context:**
When attempting to update RBS to the latest terra-common-lib, the tests failed because gson cannot serialize the project billing info:
```
com.google.gson.JsonIOException: Failed making field 'java.lang.ref.Reference#referent' accessible; either increase its visibility or write a custom TypeAdapter for its declaring type. See https://github.com/google/gson/blob/main/Troubleshooting.md#reflection-inaccessible
```

**Changes:**
Use the custom type adapter added in https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/179


samanehsan:
The action is trying to get a secret that is no longer in vault so I'm updating that now 👀

dvoet (approved):

marctalbott (approved):

PR #192, opened 2025-07-08
[CORE-593] Switch from jfrog to gar

calypsomatic:
https://broadworkbench.atlassian.net/browse/CORE-593

See https://github.com/DataBiosphere/terra-cloud-resource-lib/actions/runs/16204570168/job/45751585466?pr=192 for a successful publishing run, and https://console.cloud.google.com/artifacts/maven/dsp-artifact-registry/us-central1/libs-release-standard/bio.terra:terra-cloud-resource-lib?inv=1&invt=Ab2UUA&project=dsp-artifact-registry to see that it was successfully added to the registry.


README.md
```diff
@@ -71,44 +71,3 @@ Automatically fix linting issues:
 ./gradlew spotlessApply
 ```
 
-## Publishing an update
-
```
calypsomatic, line 75: It looks like the artifacts listed here are all very old (at least as old as the last time this list was updated, about 2 years ago), and it's not clear that they're still being published (as opposed to a single TCRL artifact), so I've removed this part of the README.

